PR #3, opened 2022-06-03
Kepler model

KaiyiLiu1234:
This is a WIP PR. I completed the basic Linear Regression Models. I added a Flask Get Request for downloading the model to Kepler. Since the Prometheus scraper has not been set up yet from Kepler, a demo will be added which will test the models on Excel Datasets using Pandas with mock data.


rootfs:
thank you!

PR #5, opened 2022-06-06
Completed Regression Models and Included Demo

KaiyiLiu1234:
In this PR, I tested and completed the basic regression models for targets Dram and Core Energy Consumption according to the desired features, and I completed saving and retraining features of said models. A demo using excel mock data was included to simulate how data from Prometheus scraper would interact with the models. Next steps include modifying the regression models for better results, ensuring requests for saved models is functional, and setting up the Prometheus scraper connection.


server/core_dataset1.csv
```diff
@@ -0,0 +1,5 @@
+39965	"Coffee Lake"	10129783825	806937020	123
```
rootfs, line 1: These are not used by production, can you move these csv to the test? It would be great when unit tests use them for validation.

rootfs:
Looks great! Please move the unit test dataset and they are ready to be merged.

KaiyiLiu1234:
In this PR, I separated Testing code from production code. I also made the code for the regression models more efficient and scalable.

PR #8, opened 2022-06-14
Refractored and improved code

KaiyiLiu1234:
Before adding the new flask route for retrieving model weights and before implementing the Prometheus based pipeline, I noticed some important code changes that can make development for future contributors easier. First change I made was adding some decoupling between the flask server and the module for training the models called kepler_model_trainer.py (code that deals with directly handling the models are now placed in the kepler_model_trainer and code that deals with handling requests are strictly in the flask server). Second change I made was ensuring the model_type (which is for accessing the different linear regression trained models) was universal. Currently, we just have two models: core energy consumption and dram energy consumpion. For example, in one module, the model_types were CoreEnergyConsumption and DramEnergyConsumption. In another module, the model_types were core_model and dram_model. When Kepler makes a get request for the desired trained model, the acceptable model_types are core_model and dram_model. Thus for simplicity, I replaced  CoreEnergyConsumpion with core_model and DramEnergyConsumption with dram_model in all relevant modules. Followup changes include minor import statement removals and removing unnecessary code. After this refactoring, I intend to then push the new flask route which allows Kepler to retrieve model weights.

### DIFF
--- a/server/kepler_model_trainer.py
+++ b/server/kepler_model_trainer.py
@@ -3,9 +3,10 @@ from keras.models import Sequential, load_model
 from keras import layers
 import numpy as np
 import os
+import shutil
 
 # Dict to map model type with filepath
-model_type_to_filepath = {"CoreEnergyConsumption": "models/core_model", "DramEnergyConsumption": "models/dram_model"}
+model_type_to_filepath = {"core_model": "models/core_model", "dram_model": "models/dram_model"}
 
 # Generates a new regression model with normalized features
 # precondition: data has already been cleaned (no missing data)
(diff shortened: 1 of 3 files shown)


server/kepler_model_trainer.py
```diff
@@ -22,12 +23,26 @@ def generate_regression_model(train_features) -> Sequential:
...
     return new_linear_model
 
+# Helper function to verify model_type is valid, check whether the model has been created or not,
+# and create a filepath to the model.
+# Returns type (str, Bool) where str is the filepath to the model_type or None if the model_type
+# is invalid, and Bool is True if the model of the desired type was already created and False if the model
+# of the desired type has not been created. If str is None, then Bool is False.
+def return_model_filepath(model_type):
+    if model_type in model_type_to_filepath:
+        filepath = os.path.join(os.path.dirname(__file__), model_type_to_filepath[model_type])
```
jeffvance, line 33: can the code the returns the model path be consolidated? See also [L82](https://github.com/sustainable-computing-io/kepler-model-server/pull/8/files#diff-fe87d5b32ed8308e36f1c74dbbc7e05e675778ecdcc3a1984e0ec3cc1abd201dR82)

PR #11, opened 2022-06-28
Kepler model

KaiyiLiu1234:
First step was to revamp the model. Used Keras Functional API and used up to date code for preprocessing. The models used are still strictly a linear regression model, and they seem to be inaccurate (but this is because the datasets only have 5 rows and are completely made up). Second step included implementing the prometheus scraper. All that is needed is to include the configuration data for the tfio.experimental.IODataset.from_prometheus which should be in the configuration yaml on Kepler. A temporary scheduler was also provided which will commence scraping and training every day at 12 am. This can be replaced with a cron job if necessary. A major todo would be to implement a function which will check the loss after training and only save the model if the loss is in an acceptable range. This should be further discussed when the kepler-model-server is tested using real metrics from kepler on AWS.

### DIFF
--- a/server/energy_scheduler.py
+++ b/server/energy_scheduler.py
@@ -1,0 +1,21 @@
+import schedule
+from prometheus_datapipeline_functions import retrieve_and_clean_prometheus_energy_metrics, create_prometheus_core_dataset, create_prometheus_dram_dataset
+from kepler_model_trainer import train_model_given_data_and_type
+import time
+
+#TODO: Test with Kepler on AWS
+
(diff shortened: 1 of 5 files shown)


KaiyiLiu1234:
Also worthy of noting, I tried to figure out how to retrieve the weights and the corresponding feature names for the coefficients. However, I was only able to return the weights which in itself is not that useful. I don't think tensorflow allows getting corresponding feature names for the coefficients even though it is a regression model (unless I am mistaken). Workarounds could include having Kepler directly using the model server for prediction or having Kepler retrieve and work with the entire trained model.

PR #19, opened 2022-07-12
schedule trainer

rootfs:


KaiyiLiu1234 (approved):
It looks like you're accounting for the delay in running energy_prometheus_pipeline which is valid because training can take a while and can possibly be more than the query interval (shown by use of time_lapsed). Also it ensures scrapes occur at same time as prometheus scrape intervals so no data is lost.

Issue #16, opened 2022-07-06
Integrate Kepler with model server


husky-parul:
As a developer, I want to use Model server to train power estimate models so that the Kepler can get regression model coefficients and use them to quickly estimate per Pod level energy consumption.

1. Model trainer to query prometheus
2. Model trainer to train power model
3. Model Server exposes coefficients
4. Kepler to retrieve coefficients
5. Kepler uses coefficients
6. Model server and training deployment manifest
7. Document architecture

sunya-ch:
- [x] Querying and online training functionality (cover 1-2) should be  done by  
https://github.com/sustainable-computing-io/kepler-model-server/blob/a13ec978d4e73fa168d7ba3cad7b9cdff2477a50/server/main.py#L142-L156
- [x] Model providing (cover 3) should be done by 
https://github.com/sustainable-computing-io/kepler-model-server/blob/a13ec978d4e73fa168d7ba3cad7b9cdff2477a50/server/model_server.py#L64-L101
- [ ] Kepler integration (cover 4-5) should be done by 
   - https://github.com/sustainable-computing-io/kepler/pull/225
   - https://github.com/sustainable-computing-io/kepler-estimator/pull/5
- [ ] (6.) Manifests and tests are in this PR https://github.com/sustainable-computing-io/kepler-model-server/pull/47    
- [ ] (7.) Check this https://github.com/sustainable-computing-io/kepler-model-server/pull/45

PR #20, opened 2022-07-12
Deployment manifests for using model trainer as a side car 

husky-parul:
Signed-off-by: Parul <parsingh@redhat.com>

Part of https://github.com/sustainable-computing-io/kepler-model-server/issues/16

This PR creates the manifest to run Kepler Model Server as a side car. The KMS queries prometheus to train/retrain power model

### DIFF
--- a/manifests/deployment.yaml
+++ b/manifests/deployment.yaml
@@ -1,3 +1,33 @@
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: kepler-model-server-pv
+  labels:
+    type: local
+    app.kubernetes.io/component: kepler-model-server-pv
+    app.kubernetes.io/name: kepler-model-server-pv
+spec:
+  storageClassName: default
+  capacity:
+    storage: 5Gi
+  accessModes:
+    - ReadWriteOnce
+  hostPath:
+    path: "/mnt/data"
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: kepler-model-server-pvc
+  namespace: monitoring
+spec:
+  storageClassName: default
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 5Gi
+---
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -7,7 +37,7 @@ metadata:
     app.kubernetes.io/component: model-server
     app.kubernetes.io/name: kepler-model-server
 spec:
-  replicas: 3
+  replicas: 1
   selector:
     matchLabels:
       app.kubernetes.io/component: model-server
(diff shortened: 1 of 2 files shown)


manifests/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+    spec:
+      containers:
+      - name: kepler-model-server
+        image: quay.io/sustainable_computing_io/kepler_model_server_base:latest
+        ports:
+        - containerPort: 8100
+...
+
+apiVersion: batch/v1
+kind: CronJob
```
rootfs, line 25: the scheduler now runs as a daemon process in #19. So we can add a sidecar container to the model server pod.

manifests/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+  replicas: 3
+  selector:
+    matchLabels:
+      app: sustainable-computing.io
+  template:
+    metadata:
+      labels:
+        app: sustainable-computing.io
+    spec:
+      containers:
```
rootfs, line 17: also add a persistent storage so that the models are persisted and can be reused.

manifests/deployment.yaml
```diff
@@ -2,39 +2,58 @@ apiVersion: apps/v1
...
   template:
     metadata:
       labels:
-        app: sustainable-computing.io
+        app.kubernetes.io/component: kepler-model-server
+        app.kubernetes.io/name: kepler-model-server
     spec:
+      volumes:
+      - name: models
+        hostPath:
```
rootfs, line 23: if possible, use a PVC

manifests/deployment.yaml
```diff
@@ -2,39 +2,58 @@ apiVersion: apps/v1
...
-metadata:
-  name: kepler-model-trainer
-spec:
-  schedule: "* * * * *"
-  jobTemplate:
-    spec:
-      containers:
+        volumeMounts:
+          - name: models
+            mountPath: /tmp/kepler-model-server/models
```
rootfs, line 32: set the mountPath to the model path defined [here](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/server/kepler_model_trainer.py#L9), ie. under the `current work dir/models`

manifests/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - name: kepler-model-server
+        image: quay.io/sustainable_computing_io/kepler_model_server_base:latest
+        ports:
+        - containerPort: 8100
+        volumeMounts:
+          - name: models
+            mountPath: /models
+      - name: kepler-model-trainer
+        image: quay.io/sustainable_computing_io/kepler_model_server_base:latest
+        command: ["python",  "energy_scheduler.py'"]
```
husky-parul, line 65: We need to add `python` to the base image `$PATH`. As of now I see this error
```
  Warning  Failed   2m34s                  kubelet  Error: container create failed: time="2022-07-18T11:54:40-04:00" level=error msg="runc create failed: unable to start container process: exec: \"python\": executable file not found in $PATH"
```

rootfs: it is installed, see [this](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/server/Dockerfile#L4)

manifests/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+        persistentVolumeClaim:
+          claimName: kepler-model-server-pvc
+      containers:
+      - name: kepler-model-server
+        image: quay.io/sustainable_computing_io/kepler_model_server_base:latest
+        ports:
+        - containerPort: 8100
+        volumeMounts:
+          - name: models
+            mountPath: /models
```
rootfs, line 62: the mountPath needs to match the work dir so the models are saved in the PV. Or change the path defined [here ](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/server/kepler_model_trainer.py#L11) as absolute path  and match the mountPath

manifests/deployment.yaml
```diff
@@ -0,0 +1,86 @@
+apiVersion: v1
+kind: PersistentVolume
```
rootfs, line 2: default SC and the provisioner should be able to create a PV for the PVC

manifests/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kepler-model-server
+  namespace: monitoring
+  labels:
+    app.kubernetes.io/component: model-server
+    app.kubernetes.io/name: kepler-model-server
+spec:
+  replicas: 3
```
rootfs, line 40: use 1 replica if using a PVC, or use a PV claim template and StatefulSet for multiple replicas.

rootfs (approved):

manifests/deployment.yaml
rootfs: please have a followup PR to use SC provisioner

PR #22, opened 2022-07-15
Finalized models, completed investigation into model weights and labe…

KaiyiLiu1234:
…ls, implemented route for model weights and labels. Kepler model server can now return a json containing important labels and corresponding weights. Keras unfortunately does not explicitly show the coefficients and corresponding feature/label name. This is likely because the coefficients provided by keras were not really intended to be used externally. However, after investigation, the order that the weights are shown by keras can be predicted by order of which the labels are fed into the model and by the order in which vocabulary is listed for categorical labels. Numerical labels also have mean and variance since all numerical labels are normalized since numerical values can get extremely large. Note that all categorical labels are one hot encoded. For instance, cpu_architecture has 8 options. Also, 'UNK' (unknown vocabulary variable) was excluded from the cpu_architecture vocabulary. Before merging this PR, consider whether 'UNK' is needed or not as vocab that is not recognized by cpu_architecture will result in an error. Also, the current vocab used by cpu architecture is "Sandy Bridge", "Ivy Bridge", "Haswell", "Broadwell", "Sky Lake", "Cascade Lake", "Coffee Lake", "Alder Lake". Please let me know whether some cpu_architectures are missing.

### DIFF
--- a/server/kepler_model_trainer.py
+++ b/server/kepler_model_trainer.py
@@ -1,6 +1,6 @@
 import tensorflow as tf
 from keras.models import Sequential, load_model
-from keras import layers, Model, Input, metrics, optimizers
+from keras import layers, Model, Input, metrics, optimizers, losses
 import numpy as np
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


KaiyiLiu1234:
Attached is a sample of the weights json returned to kepler
core_model_weights
dram_model_weights
.

KaiyiLiu1234:
I also believe it is unlikely that tensorflow will change the order the coefficients are returned for Dense layers. It wouldn't make much sense if they did.

rootfs (approved):

rootfs:
great work!

PR #31, opened 2022-07-26
Setup Prometheus /metrics endpoint for kepler model server for weights

KaiyiLiu1234:
Setup a rough Prometheus /metrics endpoint for exporting weights. In the future, would like to consider using a lightweight sql database to improve runtime.

### DIFF
--- a/server/kepler_model_trainer.py
+++ b/server/kepler_model_trainer.py
@@ -6,9 +6,10 @@ import os
 import shutil
 from keras import backend as K
 from keras.callbacks import History
+import requests
 
 # Dict to map model type with filepath
-model_type_to_filepath = {"core_model": "/models/core_model", "dram_model": "/models/dram_model"}
+model_type_to_filepath = {"core_model": "models/core_model", "dram_model": "models/dram_model"}
 
 core_model_labels = {
                 "numerical_labels": ["curr_cpu_cycles", "current_cpu_instructions", "curr_cpu_time"],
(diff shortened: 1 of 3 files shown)


server/kepler_model_trainer.py
```diff
@@ -6,9 +6,10 @@
 import shutil
 from keras import backend as K
 from keras.callbacks import History
+import requests
 
 # Dict to map model type with filepath
-model_type_to_filepath = {"core_model": "/models/core_model", "dram_model": "/models/dram_model"}
+model_type_to_filepath = {"core_model": "models/core_model", "dram_model": "models/dram_model"}
```
rootfs, line 12: use absolute paths, e.g. `/var/lib/kepler-model-server`, this should also match the mount path [here](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/manifests/deployment.yaml#L71)

KaiyiLiu1234: I can do that, but I am also using os.path.dirname(__file__) and appending it to /models/core_model which should give the absolute path of the file it currently is in no matter where. Does that not work?

rootfs: That should be fine, the path has to be deterministic so the deployment yaml and the source file can match.

KaiyiLiu1234:
I changed the path back. This created a bug for os.path.join so I replaced that. I also added the desired fixes to the model and it works  better. I will likely reduce the epochs because I noticed that after a reduction in rmse, the rmse began to increase again.

docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
+version: "3"
```
rootfs, line 1: can you make a directory for this file and other files used by docker compose, e.g. under `deploy/docker-compose`?

KaiyiLiu1234: Ok will do. I will also update the read me.

rootfs (approved):

docker-compose.yaml
rootfs: cool, looking forward to the change

PR #32, opened 2022-07-27
LR experiments notebook

Shreyanand:
Hi @KaiyiLiu1234, nice work setting up the enviornment :100:  
Ptal at the notebook in this PR, it is a starting point for experimenting with the tensorflow model and here are some other things that you can try:

1) Start small, i.e., with one feature X and see how the performance is
2) Double check the data; any unamed columns with all NaN values should be removed
3) Double check your normalization and encoding outputs (I haven't done that yet)
4) Plot the loss curve to see the training progress. With your settings I saw the curve end on a downward slope indicating the training needed more time. So, I increased the learning rate to 0.5 and epochs to 125 to reach convergence and the performance improved.
5) https://www.tensorflow.org/tutorials/keras/regression Look at the steps here and try out the Keras Sequential model version
6) Make sure your loss curve converge like it does in this notebook
7) Try DNNs present in this tutorial for an even better performance

Let me know if this is useful or if there is something still missing here.

Additionally, it can help to redefine your regression problem in a classification problem. So in this case if your goal is to predict energy consumption, you could define classes like low, medium, high or maybe 5 levels and then predict which class your system will belong to. I can help better if I know the business problem you're addressing.


Shreyanand:
Also, you could consider adding https://www.reviewnb.com/ to your repo for an easier review process of the notebooks

rootfs:
> Also, you could consider adding https://www.reviewnb.com/ to your repo for an easier review process of the notebooks

Good idea, i just added.

rootfs:
https://app.reviewnb.com/sustainable-computing-io/kepler-model-server/pull/32/

KaiyiLiu1234 (approved):

PR #33, opened 2022-09-01
Resolved docker deployment issues for prometheus and opentelemetry

KaiyiLiu1234:
Opentelemetry metrics are now working properly and are currently exported as logging. For future, a dataset like honecomb will be required. The docker deployment continuously scrapes model server's prometheus metrics and exports them as opentelemetry metrics. Dockerfiles are also now stored in a separate directoy for concision.


KaiyiLiu1234:
From what I have read, it seems the only possible way to organize the docker deployments into a single directory is to create a directory to store the docker related files and then move the server directory into said directory. Is this ok?

rootfs (approved):

rootfs:
dockerfile relocation is fine, i just updated the gh action to pick up the new path

PR #34, opened 2022-09-02
add train pipeline guideline for compatability to kepler-estimator 

sunya-ch:
This PR introduces a guideline and abstract pipeline class to build a train pipeline that is compatible to kepler-estimator and new exporting prometheus metrics.
Please check the [train_pipeline.md](https://github.com/sunya-ch/kepler-model-server/blob/train-1/doc/train_pipeline.md) for detail.

Notes: I reused the existing training function for defining a training pipeline example in [keras_pipe.py](https://github.com/sunya-ch/kepler-model-server/blob/train-1/server/train/keras_pipe.py). 

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

PR #35, opened 2022-09-02
update requirements

rootfs:
to pick up the new requirements from #34


KaiyiLiu1234 (approved):
For requirements.txt, numpy needs to be included. I can include that since it's a quick fix.

PR #37, opened 2022-09-09
functions to configurable parameters and shared paths

sunya-ch:
This PR implements functions to handle two environment configurations.

1. configurable parameters
I think we should set model parameters from config map if we are using Kubernetes-based cluster.
A key benefit is that we can share a single point of configuration. For example, share prometheus server endpoint/port with the co-operated prometheus system. share model path with Kepler estimator container.
The `getConfig` will support both setting from configmap mount volume and environment variable.

2. shared paths
I think we can provide model via the shared file system such as NFS.
This will prepare model path (create if not exists) at specific mount point. This is also extensible to other path such as initial dataset.

### DIFF
--- a/server/util/config.py
+++ b/server/util/config.py
@@ -1,0 +1,44 @@
+#################################################
+# config.py
+#
+# getConfig: return value set by configuration 
+#            which can be from config map or environment variable
+#            if not provided, return default value
+# getPath:   return path relative to mount path
+#            create new if not exists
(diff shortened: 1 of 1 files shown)


rootfs:
thanks for the ideas. I think in the future the models are going to be stored in object storage as well (in addition to NFS).

rootfs (approved):

PR #38, opened 2022-09-09
refactor prom query

sunya-ch:
This PR refactors the prometheus query approach with the following implementations.

- Instead of manually specify the metric label key on every reference, this refactoring  use pandas from handling query data and use a single point of grouped feature definition in `train/train_types.py`.
- All queries and prometheus parameters are from the configuration.
- Prefix `node_`, `curr_` are removed from the label to share grouped feature definition between pod and node stat.
- Query multiple metrics at the same time because the different trainning pipelines may use different set of query. For example, dynamic power trainning pipeline use pod stat not node stat.

note: there is a dependent getConfig call from PR https://github.com/sustainable-computing-io/kepler-model-server/pull/37.

usage:
```python
# call only every interval in energy_scheduler
prom_client.query()

# call for each training pipieline (used the same cached query); 
# return all available columns if features=None
data = prom_client.get_data(query_metric, features)
```

### DIFF
--- a/server/dockerfiles/requirements.txt
+++ b/server/dockerfiles/requirements.txt
@@ -5,3 +5,4 @@ protobuf==3.19.4
 prometheus-client==0.14.1
 pandas==1.4.4
 numpy==1.22.4
+prometheus-api-client==0.5.1
--- a/server/prom/query.py
+++ b/server/prom/query.py
@@ -1,0 +1,91 @@
+import os
+import sys
+
+server_path = os.path.join(os.path.dirname(__file__), '..')
+sys.path.append(server_path)
+
+from prometheus_api_client import PrometheusConnect
+import datetime
+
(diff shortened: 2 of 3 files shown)


KaiyiLiu1234:
Is there a specific reason why you chose to save as h5 instead of as a SavedModel? Recommended way of saving is with SavedModel. This is not relevant to these PRs but it was something I noticed looking through the changes.

sunya-ch:
@KaiyiLiu1234 Thank you for the review. No particular reason to use `.h5` except I had some issue with version compatibility before. If we both confirm the version on both kepler-estimator and kepler-model-server, we can also use the SavedModel instead. I just haven't checked that yet.

KaiyiLiu1234:
No worries. Can further discuss in upcoming meeting.

PR #40, opened 2022-09-12
correct env settings in query.py

sunya-ch:
This PR fixes typo (PROM_SERVER) from my previous PR and corrects the assignments of PROM_SSL_DISABLE, PROM_HEADERS converting from string. 


server/prom/query.py
```diff
@@ -72,10 +72,11 @@ def query(self):
             df = pd.DataFrame(items) 
             df.columns = df.columns.str.replace("curr_", "")
             df.columns = df.columns.str.replace("node_", "")
-            df[query_metric] = df['value']
-            for col in df.columns:
-                df[col] = df[col].transform(transform_float)
-            df.drop(columns=['value'], inplace=True)
+            if len(df) > 0:
```
KaiyiLiu1234, line 75: Is it possible for df to be nan or is it always an empty string? Might need to consider pandas isnull function.

KaiyiLiu1234 (approved):

server/prom/query.py
sunya-ch: df will be always initialized by `pd.DataFrame(items) `. with no data, items = []. and df will just has no data

PR #44, opened 2022-09-19
Merging PR 42 changes

KaiyiLiu1234:


KaiyiLiu1234:
Also merged energy scheduler with main.py For future reference, maybe it would be good to keep deprecated code somehow (either labeling them as deprecated with python's deprecated or some other method). Could be useful for reference and learning.

sunya-ch:
> Also merged energy scheduler with main.py For future reference, maybe it would be good to keep deprecated code somehow (either labeling them as deprecated with python's deprecated or some other method). Could be useful for reference and learning. Any thoughts on this?

I'm not sure I understand it correctly. We are going to replace code energy_scheduler.py with main.py right? If we replace it the previous code should be in the commit history I think. Also, we can create a new branch from the point before merge to keep all the previous version.

KaiyiLiu1234:
The first commits are yours. I believe there was a merge conflict. Yes we are going to replace energy scheduler with main (that has been done). I was thinking for future to consider deprecating when users start using functions we don't want them to. But you're right those other options are good as well.

server/main.py
```diff
@@ -34,10 +34,52 @@
...
+    train_model_given_data_and_type(core_train, core_val, core_test, 'core_model')
+    train_model_given_data_and_type(dram_train, dram_val, dram_test, 'dram_model')
+
+# scrape prometheus and dummy test metrics
+def dummy_prometheus_pipeline(query, interval, endpoint):
+    test_dataset = retrieve_and_clean_prometheus_energy_metrics(query, interval, endpoint)
+    print(test_dataset)
+
+
+def run_energy_scheduler_pipeline():
```
sunya-ch, line 112: should be moved to deprecated branch later

sunya-ch (approved):

PR #67, opened 2023-01-10
Update Features

KaiyiLiu1234:
This PR includes updating feature names and including CPU Frequency. An Issue I wanted clarification on was what are the metric columns for kepler_node_cpu_scaling_frequency_hertz and kepler_container_cpu_cpu_time_us queries to be used as features in node training? It's not very clear what the metric name should be. Refer: https://www.notion.so/Sync-up-Kepler-version-b79303ddc3814308aa9f134b1031c261


sunya-ch:
For node training, we may need cpu_frequency feature. `kepler_node_cpu_scaling_frequency_hertz` provides frequency per cpu core. We may simply find stats such as min, max, average, std and use and features. Optionally, we can also use the information of usage on each cpu core from `kepler_container_cpu_cpu_time_us` to give a weight of frequency for each core too (instead of equally taking an average). We may try multiple ways to generate the features to find the best model.

PR #80, opened 2023-02-21
Completed XGBoost Pipeline with demo code

KaiyiLiu1234:
Contains the completed XGBoost Pipeline for package power for Standalone IRQ metrics and Cgroup.


sunya-ch (approved):
/lgtm

server/train/train_types.py
```diff
@@ -10,14 +10,16 @@
...
 
 SYSTEM_FEATURES = ["cpu_architecture"]
 
 COUNTER_FEAUTRES = ["cache_miss", "cpu_cycles", "cpu_instr"]
 CGROUP_FEATURES = ["cgroupfs_cpu_usage_us", "cgroupfs_memory_usage_bytes", "cgroupfs_system_cpu_usage_us", "cgroupfs_user_cpu_usage_us"]
 IO_FEATURES = ["bytes_read", "bytes_writes"]
 BPF_FEATURES = ["cpu_time"]
-KUBELET_FEATURES =['kubelet_memory_bytes', 'kubelet_cpu_usage']
+KUBELET_FEATURES = ['kubelet_memory_bytes', 'kubelet_cpu_usage']
+IRQ_FEATURES = ["block_irq", "net_tx_irq", "net_rx_irq"]
```
rootfs, line 22: since `cpu_time` and all irq metrics are all from ebpf program, we can consolidate them into the new `BPF_FEATURES`

server/train/pipe_util.py
```diff
@@ -248,3 +259,214 @@ def merge_model(core_model, dram_model):
...
+        self.save_location = save_location
+        self.model_name = model_name
+        self.model_filename = model_name + '.model'
+        self.model_desc = 'model_desc.json'
+
+
+    @staticmethod
+    def _generate_base_model() -> xgb.XGBRegressor:
+        # n_estimators, max_depth, eta (learning rate)
+        return xgb.XGBRegressor(n_estimators=1000, learning_rate=0.1)
```
rootfs, line 301: if curious, is there any hyperparaemeter tuning in xgboost package, like `GridSearchCV` and `RandomizedSearchCV` in sklearn?

server/train/train_types.py
KaiyiLiu1234: Ok I will remove IRQ Features and instead relocate these metrics into BPF Features

server/train/pipe_util.py
KaiyiLiu1234: There is hyperparameter tuning because xgboost does use tools from sklearn. If I recall, we need to provide a list of candidate values for key parameters like n_estimators, max_depth, learning rate, etc. While I am not too sure, I think this can be incorporated in the online training. I will look into it. I will include this in the next PR.

PR #87, opened 2023-02-27
Update Standalone

KaiyiLiu1234:
Updated standalone and dyncomp model to develop models for node and container level. Updated predict for standalone model to accept a dict containing feature names and corresponding values. Updated extractor_test.py to contain additional examples. Updated train_types to include enums for Label groups instead of only feature groups.

### DIFF
--- a/server/train/pipelines/XGBoostRegressionCompDynPipeline/pipe.py
+++ b/server/train/pipelines/XGBoostRegressionCompDynPipeline/pipe.py
@@ -2,40 +2,47 @@ import os
 import sys
 from typing import List, Tuple, Dict, Any
 import pandas as pd
+import copy
 
 train_path = os.path.join(os.path.dirname(__file__), '../../')
 prom_path = os.path.join(os.path.dirname(__file__), '../../../prom')
 sys.path.append(train_path)
 sys.path.append(prom_path)
 
-from pipeline import TrainPipeline
-from train_types import FeatureGroup, FeatureGroups, PACKAGE_LABEL, ModelOutputType, XGBoostRegressionTrainType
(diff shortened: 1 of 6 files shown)


KaiyiLiu1234:
By pushing this, I should also be able to test on kind cluster. kepler-model-server-base already installed the sklearn dependency. By making a second push, kepler-model-server should then also install the sklearn dependency from kepler-model-server-base (please let me know if this is incorrect assumption). Also, for the next PR which will be pushed today, I plan to update extractor.py to accept the new kepler metrics. The following PR will then be focused on setting up the dyn-comp pipeline to handle multiple models (for package power and dram power). How this is implemented will depend on @sunya-ch implementation of the new pipeline api.

server/train/pipelines/XGBoostRegressionCompDynPipeline/pipe.py
```diff
@@ -2,40 +2,47 @@
...
-        model_class = 'xgboost'
+        self.train_type = train_type
+        self.model_class = 'xgboost'
         self.energy_source = 'rapl'
-        self.energy_components = ['package', 'dram']
-        features = FeatureGroups[FeatureGroup.CgroupOnly]
-        model_output = ModelOutputType.DynComponentPower
-        self.labels = PACKAGE_LABEL
-        super().__init__(model_name, model_class, model_file, features, model_output)
+        self.feature_group = FeatureGroup.CounterOnly
```
rootfs, line 26: also include irq features

KaiyiLiu1234: Included. I should note that in the next PR (which involves updating extractor for kepler metrics), IRQ features will be stored in BPF. Moreover, pipeline won't work yet because the kepler metrics have not been updated in the kepler model server. Will finish this today.

rootfs (approved):

PR #88, opened 2023-02-28
Add new training pipeline source code with test

sunya-ch:
This PR initially completes the implementation of new pipeline stated in https://github.com/sustainable-computing-io/kepler-model-server/issues/75.

I put new pipeline implementation separately in the new folder to avoid conflict with the current implementation. 
However, when we fully migrates the previous trainer to new pipeline, the old server folder must be replaced.

We can try testing each component of the pipeline as written in workflow action.

```yaml
    - name: Test extractor
      run: |
         python3 tests/extractor_test.py

    - name: Test isolator
      run: |
         python3 tests/isolator_test.py

    - name: Test trainer
      run: |
         python3 tests/trainer_test.py

    - name: Test pipeline
      run: |
         python3 tests/pipeline_test.py
```

The output of `python3 tests/pipeline_test.py` will be in `src/models` folder.
```bash
src/models
├── acpi
│   ├── AbsPower
│   │   ├── BPFOnly
│   │   │   ├── GradientBoostingRegressorTrainer_1
│   │   │   │   ├── GradientBoostingRegressorTrainer_1.json
│   │   │   │   ├── metadata.json
│   │   │   │   ├── platform.pkl
│   │   │   │   └── scaler.pkl
│   │   │   ├── GradientBoostingRegressorTrainer_1.zip
│   │   │   ├── SGDRegressorTrainer_1
│   │   │   │   ├── SGDRegressorTrainer_1.json
│   │   │   │   ├── metadata.json
│   │   │   │   ├── platform.pkl
│   │   │   │   ├── scaler.pkl
│   │   │   │   └── weight.json
│   │   │   ├── SGDRegressorTrainer_1.zip
│   │   │   └── checkpoint
│   │   │       ├── GradientBoostingRegressorTrainer_platform_1.pkl
│   │   │       └── SGDRegressorTrainer_platform_1.pkl
...
│   └── DynPower
│       ├── BPFOnly
│       │   ├── GradientBoostingRegressorTrainer_1
│       │   │   ├── GradientBoostingRegressorTrainer_1.json
│       │   │   ├── metadata.json
│       │   │   ├── platform.pkl
│       │   │   └── scaler.pkl
│       │   ├── GradientBoostingRegressorTrainer_1.zip
│       │   ├── SGDRegressorTrainer_1
│       │   │   ├── SGDRegressorTrainer_1.json
│       │   │   ├── metadata.json
│       │   │   ├── platform.pkl
│       │   │   ├── scaler.pkl
[...]

Issue #75, opened 2023-02-09
New online training pipeline (extractor-->(isolator)-->trainer)


sunya-ch:
Unlike node power, workload (pod/container/process) power should not take idle and background power into the model. Otherwise, the consumed power aggregated from all workload will be over the actual measured power. 
The idle and background power should be identified and removed from the measured power before training. 

So, I proposed a new pipeline to perform online training as below.

![online_trainer_removed](https://user-images.githubusercontent.com/11749848/217743609-f72659f6-25b1-41e4-bb7e-2b64d10e0857.png)

- [x] extractor by https://github.com/sustainable-computing-io/kepler-model-server/pull/77
- [x] isolator - idle power
- [x] isolator - background power
- [x] trainer 
- [x] combine pipeline

Signed-off-by: Sunyanan Choochotkaew <sunyanan.choochotkaew1@ibm.com>

sunya-ch:
WIP: https://github.com/sunya-ch/kepler-model-server/tree/pipeline/server/train


```python
class Pipeline():
    def __init__(self, trainers, extractor=DefaultExtractor(), isolator=MinIdleIsolator()):
        self.extractor = extractor
        self.isolator = isolator
        self.trainers = trainers

    def process(self, query_results, energy_components, feature_group, energy_source, node_level, aggr=True):
        extracted_data, power_labels = self.extractor.extract(query_results, energy_components, feature_group, energy_source, node_level, aggr)
        if not node_level:
            isolated_data = self.isolator.isolate(extracted_data, power_labels)
        for trainer in self.trainers:
            trainer.train(isolated_data)
```

sunya-ch:
Let assume we have a model to cluster node_type. 

Still thinking about end-to-end flow.
- how profile generated  
   - currently, manual run script in https://github.com/sunya-ch/kepler-model-server/tree/pipeline/profile
   - can we allow PR to upload dataset and then automatically apply the script then remove the dataset with CI?
- how and when model will be updated 
   - currently, it's always updated. need to properly implement should_archive function with some validation function to make sure the model is improved.
   - may allow PR to upload new model with some validation by CI.

sunya-ch:
Currently, I made the starting endpoint by providing a tool and script to let people (contributor) run the benchmark and submit to the system.
https://github.com/sunya-ch/kepler-model-server/tree/pipeline/src/profile

Example output:
```bash
> ./script.sh run
namespace/cpe-operator-system created
[...]

PR #90, opened 2023-02-28
Update server api to use new pipeline model

sunya-ch:
This PR extended and updated the server-api to work with new pipeline models (https://github.com/sustainable-computing-io/kepler-model-server/issues/75).

- [x] There are some dependencies in PR https://github.com/sustainable-computing-io/kepler-model-server/pull/88. 
Need to merge that PR first. 

Signed-off-by: Sunyanan Choochotkaew [sunyanan.choochotkaew1@ibm.com](mailto:sunyanan.choochotkaew1@ibm.com)

### DIFF
--- a/src/server/model_server.py
+++ b/src/server/model_server.py
@@ -1,0 +1,168 @@
+from flask import Flask, request, json, make_response, send_file
+
+import os
+import sys
+src_path = os.path.join(os.path.dirname(__file__), '..')
+util_path = os.path.join(os.path.dirname(__file__), 'util')
+train_path = os.path.join(os.path.dirname(__file__), '..', 'train')
+
+sys.path.append(src_path)
+sys.path.append(util_path)
+sys.path.append(train_path)
+
+from train import get_valid_feature_groups, is_weight_output, ModelOutputType, FeatureGroups, FeatureGroup
+from util.config import getConfig
+from util.loader import parse_filters, is_valid_model, load_json, get_model_weight, get_model_group_path, get_archived_file, METADATA_FILENAME, CHECKPOINT_FOLDERNAME
+
+###############################################
+# model request 
+
+class ModelRequest():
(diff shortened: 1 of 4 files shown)


src/server/model_server.py
```diff
@@ -0,0 +1,163 @@
...
+
+app = Flask(__name__)
+
+@app.route('/model', methods=['POST'])
+def get_model():
+    model_request = request.get_json()
+    req = ModelRequest(**model_request)
+    valid_fgs = get_valid_feature_groups(req.metrics)
+    filters = parse_filters(req.filter)
+    print(req.output_type)
```
KaiyiLiu1234, line 78: A few of these print statements may not be necessary and should be removed or logged for future reference.

src/server/model_server.py
```diff
@@ -0,0 +1,163 @@
...
+
+from train import get_valid_feature_groups, is_weight_output, ModelOutputType, FeatureGroups, FeatureGroup
+from util.config import getConfig
+from util.loader import parse_filters, is_valid_model, load_json, get_model_weight, get_model_group_path, get_archived_file, METADATA_FILENAME, CHECKPOINT_FOLDERNAME
+
+###############################################
+# model request 
+
+class ModelRequest():
+    def __init__(self, metrics, output_type, source='rapl', node_type=-1, weight=False, trainer_name="", filter=""):
```
KaiyiLiu1234, line 21: For future reference, might be a good idea to include comments and class descriptors just to make it easier for new developers to onboard

sunya-ch: Thanks. Added comments for each attribute.
https://github.com/sustainable-computing-io/kepler-model-server/compare/a989d7051b8056261c5df86a5594af3704517201..04aae778b8a000ff5209aa4d7008143d2262aa96

src/server/model_server.py
sunya-ch: Change from printing to comment. Could we put changing to logging as a new issue and PR applying to other component (such as online trainer) as well?

PR #95, opened 2023-04-05
change config path to match kepler exporter

sunya-ch:
Related to path change on kepler exporter in PR https://github.com/sustainable-computing-io/kepler/pull/566, there is also a path change defined in the model server.

This PR made a path change in corresponding test manifests and the source code to read the config from the expecting path. 

### DIFF
--- a/manifests/dev/deployment.yaml
+++ b/manifests/dev/deployment.yaml
@@ -48,7 +48,7 @@ spec:
         - containerPort: 8100
         volumeMounts:
           - name: cfm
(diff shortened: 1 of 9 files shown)


KaiyiLiu1234:
Thanks for the update. Will update the mountPath for the operator. Note that the Operator will for now deploy using host-local manifests (which includes the PV and PVC). Will change this in the future.

KaiyiLiu1234 (approved):
LGTM

PR #108, opened 2023-06-26
Updated Evaluation Metrics with Results

KaiyiLiu1234:
Using the test data alone, the xgboost model performs quite well and should become one of the primary models. For instance with container level IRQ metrics + CPU_Time with KFold Cross Validation, we get an average Mean Absolute Percentage Error (consider this like accuracy but the lower the better) of 8.3% (anything below 10% is very good) and average R2 of 0.5565. For Nodel Level, we get an average Mean Absolute Percentage Error of 7.4% and an average R2 of 0.735. Note we use "average" because KFold Cross Validation involves taking numerous slices for testing and training. For TrainTestSplit, there is a lot more variation so some scenarios we see R2 of 0.2 but this can be biased. If you have any questions related to KFold Cross Validation vs TrainTestSplit let me know.


KaiyiLiu1234:
@rootfs you can also use these evaluation metrics to describe the overall success of the model server. XGBoost performance is far superior compared to Linear Regression. If I recall, the R2 of the Linear Regression model was at best 0.02 if not negative and the Mean Absolute Error was greater than the average y value (so you can expect the Mean Absolute Percentage Error to be very large i.e 1 or greater)

KaiyiLiu1234:
@rootfs Note that we still need a final validation test because we want to see how the model can perform on unseen data. If we can include this final validation test which I believe is still being discussed, the model server or at least XGBoost model should have good coverage.

rootfs (approved):

PR #119, opened 2023-08-17
fix bugs estimator has no energy source

sunya-ch:
This PR additionally fixes multiple bugs due to integration test to Kepler.
- feature data is missing when there is no unit values such as in acpi platform energy
- energy source field is missing from estimator PowerRequest
- model path is set `rapl` to default which can be misleading.
- cmd train supported only single energy source
- initUrlKeyMap and estimatorKeyMap are not valid anymore. The model specification is now combination of output type and energy source. 


sunya-ch:
This PR needs retest again after this PR https://github.com/sustainable-computing-io/kepler-model-db/pull/8 merged due to missing metadata issue.

rootfs:
unit test passed with model db changes.

Issue #122, opened 2023-08-21
apply smooth in preprocessing (extractor)


sunya-ch:
### What would you like to be added?

Smoothing the input data can increase an accuracy. 

Here are my preliminary test results:

**DefaultExtractor**

model_name | model_class | model_file | features | fe_files | output_type | mae | package_mae | core_mae | uncore_mae | dram_mae | node_type
| -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- 
PolynomialRegressionTrainer_1 | scikit | PolynomialRegressionTrainer_1.json | [kubelet_cpu_usage, kubelet_memory_bytes] | [poly_scaler.pkl] | AbsPower | 18.032359 | 18.032359 | 0.0 | 0.0 | 2.888742 | 1
GradientBoostingRegressorTrainer_1 | scikit | GradientBoostingRegressorTrainer_1.json | [kubelet_cpu_usage, kubelet_memory_bytes] | [] | AbsPower | 15.042512 | 15.042512 | 0.0 | 0.0 | 4.601480 | 1
SGDRegressorTrainer_1 | scikit | SGDRegressorTrainer_1.json | [kubelet_cpu_usage, kubelet_memory_bytes] | [] | AbsPower | 877824.070562 | 877824.070562 | 0.0 | 0.0 | 3.789950 | 1
KNeighborsRegressorTrainer_1 | scikit | KNeighborsRegressorTrainer_1.json | [kubelet_cpu_usage, kubelet_memory_bytes] | [] | AbsPower | 14.434462 | 14.434462 | 0.0 | 0.0 | 4.739162 | 1
LinearRegressionTrainer_1 | scikit | LinearRegressionTrainer_1.json | [kubelet_cpu_usage, kubelet_memory_bytes] | [] | AbsPower | 20.061682 | 20.061682 | 0.0 | 0.0 | 2.915489 | 1
SVRRegressorTrainer_1 | scikit | SVRRegressorTrainer_1.json | [kubelet_cpu_usage, kubelet_memory_bytes] | [] | AbsPower | 21.491037 | 21.491037 | 0.0 | 0.0 | 3.220217 | 1

**SmoothExtractor** (tentative name)

model_name | model_class | model_file | features | fe_files | output_type | mae | package_mae | core_mae | uncore_mae | dram_mae | node_type 
| -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- 
[...]

PR #123, opened 2023-08-21
add SmoothExtractor

sunya-ch:
According to feature request issue in https://github.com/sustainable-computing-io/kepler-model-server/issues/122, this PR implements SmoothExtractor to smooth the data with specific window size on extraction. 
This is also available to test via the entrypoint `cmd/main.py` with `--extractor smooth`.

As this PR has a base on https://github.com/sustainable-computing-io/kepler-model-server/pull/121 fix, please merge the fix first.

### DIFF
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -22,6 +22,8 @@ optional arguments:
                         Specify metrix prefix to filter.
   -p PIPELINE_NAME, --pipeline-name PIPELINE_NAME
                         Specify pipeline name.
+  --extractor EXTRACTOR
+                        Specify extractor name (default, smooth).
   --isolator ISOLATOR   Specify isolator name (none, min, profile, trainer).
   --profile PROFILE     Specify profile input (required for trainer and profile isolator).
   -e ENERGY_SOURCE, --energy-source ENERGY_SOURCE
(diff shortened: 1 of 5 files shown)


src/train/trainer/__init__.py
```diff
@@ -233,7 +236,7 @@ def save_model_and_metadata(self, node_type, X_test, y_test):
         mae_map = dict()
         item = self.get_basic_metadata(node_type)
         for component in self.energy_components:
-            mae = self.get_mae(node_type, component, X_test, y_test)
+            mae = self.get_mae(node_type, component, X_test_map[component], y_test_map[component])
```
rootfs, line 239: is this still needed after #121?

rootfs:
does kepler also need to smooth the feature during inference?

sunya-ch:
> does kepler also need to smooth the feature during inference?

I don't think so because we return one-to-one prediction without historical data for the current power model.

src/train/trainer/__init__.py
sunya-ch: I will rebase the PR.

sunya-ch: Done rebasing. Please confirm it. Thanks.

rootfs (approved):

PR #135, opened 2023-08-30
Add accelerator feature

Yanbo0101:
This PR is to add accelerator feature to model training.
Now kepler can collect accelletor metrics and report to Prometheus,  for example:

kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev0",source="intel-qat",type="latency"} 46750
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev1",source="intel-qat",type="latency"} 0

And in this pr, hope to add accelerator feature(AcceleratorOnly) for querying data and training power models.

### DIFF
--- a/src/train/extractor/extractor.py
+++ b/src/train/extractor/extractor.py
@@ -13,7 +13,7 @@ from prom_types import TIMESTAMP_COL, SOURCE_COL, get_energy_unit, \
         energy_component_to_query, feature_to_query, \
             pkg_id_column, container_id_cols, node_info_column
 from loader import default_node_type
-from extract_types import container_id_colname, ratio_to_col, component_to_col, get_unit_vals
+from extract_types import container_id_colname, ratio_to_col, component_to_col, get_unit_vals, accelerator_type_colname
 from preprocess import drop_zero_column, find_correlations
 
 # append ratio for each unit
@@ -55,6 +55,7 @@ class Extractor(metaclass=ABCMeta):
     # - feature_power_data: dataframe of feature columns concat with power columns
     # - power_columns: identify power columns (labels)
     # - corr: correlation matrix between features and powers
+    # - features: updated features
     @abstractmethod
     def extract(self, query_results, feature_group):
         return NotImplemented
@@ -70,17 +71,23 @@ class DefaultExtractor(Extractor):
         # 1. compute energy different per timestamp and concat all energy component and unit
         power_data = self.get_power_data(query_results, energy_components, energy_source)
         if power_data is None:
-            return None, None, None
+            return None, None, None, None
         power_data = drop_zero_column(power_data, power_data.columns)
         power_columns = power_data.columns
-        features = FeatureGroups[FeatureGroup[feature_group]]
+        fg = FeatureGroup[feature_group]
+        features = FeatureGroups[fg]
         # 2. separate workload and system 
         workload_features = [feature for feature in features if feature not in SYSTEM_FEATURES]
         system_features = [feature for feature in features if feature in SYSTEM_FEATURES]
         # 3. compute aggregated utilization different per timestamp and concat them
-        feature_data = self.get_workload_feature_data(query_results, workload_features)
+        if fg == FeatureGroup.AcceleratorOnly and node_level is not True:
+            return None, None, None, None
+        else:
+            feature_data, workload_features = self.get_workload_feature_data(query_results, workload_features)
+
         if feature_data is None:
-            return None, None, None
+            return None, None, None, None
+
         # join power
         feature_power_data = feature_data.set_index(TIMESTAMP_COL).join(power_data).sort_index().dropna()
         
@@ -108,16 +115,18 @@ class DefaultExtractor(Extractor):
         if node_info_column not in feature_power_data.columns:
             feature_power_data[node_info_column] = int(default_node_type)
         
-        
         # 6. validate input with correlation
         corr = find_correlations(energy_source, feature_power_data, power_columns, workload_features)
         # 7. apply utilization ratio to each power unit because the power unit is summation of all container utilization
         feature_power_data = append_ratio_for_pkg(feature_power_data, is_aggr, query_results, power_columns)
-        return  feature_power_data, power_columns, corr
+        return  feature_power_data, power_columns, corr, workload_features
     
     def get_workload_feature_data(self, query_results, features):
         feature_data = None
         container_df_map = dict()
+        accelerator_df_list=[]
+        cur_accelerator_features = []
+        feature_to_remove =[]
         for feature in features:
             query = feature_to_query(feature)
             if query not in query_results:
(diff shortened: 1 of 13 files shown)


sunya-ch:
Could you also create PR related to the new metric added to the Kepler document too?
https://github.com/sustainable-computing-io/kepler-doc/blob/main/docs/kepler_model_server/pipeline.md

Yanbo0101:
> Could you also create PR related to the new metric added to the Kepler document too? https://github.com/sustainable-computing-io/kepler-doc/blob/main/docs/kepler_model_server/pipeline.md

Thanks for your comments! I will do that later.

src/train/extractor/extractor.py
```diff
@@ -177,6 +201,31 @@ def get_system_feature_data(self, query_results, features):
...
+                new_colname = "{}_{}".format(feature, type_name)
+                type_data.rename(columns={query: new_colname}, inplace=True)
+                type_data = type_data [[TIMESTAMP_COL, new_colname]]
+                type_data = type_data .groupby([TIMESTAMP_COL]).sum().sort_index()
+                tmp_list += [type_data ]
+                cur_accelerator_features.append(new_colname)
+            tmp_data_list = pd.concat(tmp_list,axis=1).fillna(0).astype(int).sort_values(by=TIMESTAMP_COL)
+            feature_data_list += [tmp_data_list]
+        feature_data = pd.concat(feature_data_list, axis=1).astype(int)
+        return feature_data, cur_accelerator_features
```
sunya-ch, line 228: Why we need to separate `accelerator_features` from other feature_data?

src/train/pipeline.py
```diff
@@ -13,7 +13,7 @@
 from extractor import DefaultExtractor
 from isolator import MinIdleIsolator
 
-from train_types import PowerSourceMap, FeatureGroups, ModelOutputType
+from train_types import PowerSourceMap, FeatureGroups, ModelOutputType, ACCELERATE_FEATURES
```
sunya-ch, line 16: ACCELERATE_FEATURES should be referred under FeatureGroup (same as other metric group).

sunya-ch (changes requested):

src/train/extractor/extractor.py
Yanbo0101: It's node level. And in kepler now, the metrics is : 
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev0",source="intel-qat",type="latency"} 46750
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev1",source="intel-qat",type="latency"} 0
May will add more like:
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev0",source="intel-qat",type="compress"} 4
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev1",source="intel-qat",type="compress"} 1
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev0",source="intel-qat",type="decompress"} 4
kepler_node_accelerator_intel_qat{instance="cpio-5",qatDevID="qat_dev1",source="intel-qat",type="decompress"} 1
Just different types, so I also need to distinguish between different types

src/train/pipeline.py
Yanbo0101: Thanks! I will fix it.

Yanbo0101: Fixed all! Thanks!

src/train/extractor/extractor.py
sunya-ch: I see. So, it is the metric that is available only on the node level. 

I think instead of separating it like this. We should modify the extractor to add a separated function to extract the node-level metrics which is not limited to only this `accelerator_features`. Also, we should support extracting multiple metrics from a single query like this case in the future. This could be a refactoring the later PR. 

For this PR, 
1. we need to add condition `node_level==True` to check whether the accelerator_features could be applied.
2. You can set `fg = FeatureGroup[feature_group]`, then check `fg == FeatureGroup.AcceleratorOnly` directly. 
3. Should define feature name per column. Not separate it to `cur_accelerator_features`.  (please remove additional return from extract function)

      ```python
      ACCELERATE_FEATURES = ["accelerator_intel_qat_latency", "accelerator_intel_qat_compress", "accelerator_intel_qat"_decompress"]
      ```
      
4. rebase from the main branch

sunya-ch (changes requested):

src/train/extractor/extractor.py
```diff
@@ -69,27 +70,43 @@ def extract(self, query_results, energy_components, feature_group, energy_source
         # 1. compute energy different per timestamp and concat all energy component and unit
         power_data = self.get_power_data(query_results, energy_components, energy_source)
         if power_data is None:
-            return None, None, None
+            return None, None, None, None
         power_data = drop_zero_column(power_data, power_data.columns)
         power_columns = power_data.columns
         features = FeatureGroups[FeatureGroup[feature_group]]
```
sunya-ch, line 76:
```suggestion
fg = FeatureGroup[feature_group]
features = FeatureGroups[fg] <--  ACCELERATE_FEATURES  must be = ["accelerator_intel_qat_latency", "accelerator_intel_qat_compress", "accelerator_intel_qat"_decompress"]

if node_level == True and fg == FeatureGroup.AcceleratorOnly:
     feature_data = self.get_accelerator_feature_data(query_results, workload_features) <-- reset_index inside this function
else:
     feature_data = self.get_workload_feature_data(query_results, workload_features)

if feature_data is None:
   return None, None, None
  

```

src/train/extractor/extractor.py
```diff
@@ -69,27 +70,43 @@ def extract(self, query_results, energy_components, feature_group, energy_source
...
         # 2. separate workload and system 
         workload_features = [feature for feature in features if feature not in SYSTEM_FEATURES]
         system_features = [feature for feature in features if feature in SYSTEM_FEATURES]
+        # 2.1 separate accelerator workload form workloads
+        accelerator_features = [feature for feature in features if feature in FeatureGroups[FeatureGroup.AcceleratorOnly]]
```
sunya-ch, line 81: No need for these two lines (81-82)

src/train/extractor/extractor.py
```diff
@@ -69,27 +70,43 @@ def extract(self, query_results, energy_components, feature_group, energy_source
...
+                return None, None, None,None
+
         # join power
         feature_power_data = feature_data.set_index(TIMESTAMP_COL).join(power_data).sort_index().dropna()
         
         # aggregate data if needed
         is_aggr = node_level and aggr
         if is_aggr:
             # sum stat of all containers
-            sum_feature = feature_power_data.groupby([TIMESTAMP_COL]).sum()[workload_features]
```
sunya-ch, line 90: Keep using the `workload_features`. FeatureGroup must be redefined for each feature column at the beginning.

src/train/extractor/extractor.py
```diff
@@ -99,6 +116,13 @@ def extract(self, query_results, energy_components, feature_group, energy_source
             feature_power_data = feature_power_data.join(system_feature_data).sort_index().dropna()
         else: 
             feature_power_data = feature_power_data.sort_index()
+        
+        # 4.1 add accelerator features (non aggregated data)
+        if accelerator_flag is not True and len(accelerator_features) > 0:
```
sunya-ch, line 121: Should not add logic here. It is only for the system features which is equal to all container. But, accelerator_feature_data is numerical that should be available only in node-level.

src/train/extractor/smooth_extractor.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, smooth_window=30):
 
     # implement extract function
     def extract(self, query_results, energy_components, feature_group, energy_source, node_level, aggr=True):
-        feature_power_data, power_columns, _ = super().extract(query_results, energy_components, feature_group, energy_source, node_level, aggr)
+        feature_power_data, power_columns, _, accelerator_features = super().extract(query_results, energy_components, feature_group, energy_source, node_level, aggr)
```
sunya-ch, line 17: remove need of the extra return.

src/train/extractor/smooth_extractor.py
```diff
@@ -26,5 +26,5 @@ def extract(self, query_results, energy_components, feature_group, energy_source
 
         corr = find_correlations(energy_source, feature_power_data, power_columns, workload_features)
 
-        return smoothed_data, power_columns, corr
+        return smoothed_data, power_columns, corr, accelerator_features
```
sunya-ch, line 29: remove need to extra return.

sunya-ch (commented):
We may need to discuss again later about node-level metrics. At this step, let have it available only for the AbsPower model.

src/train/extractor/extractor.py
Yanbo0101: Thanks for your comments. How about not judging whether node_ is equal to true now? 
If node_level == False, get_workload_feature_data() will handle accelerator_features, or directly return None, None, None，which may cause other errors.

src/train/extractor/extractor.py
Yanbo0101: Actually，when fg==FeatureGroup.WorkloadOnly, I still need to separate the accelerator workload from the workloads.

src/train/extractor/extractor.py
sunya-ch: You are right. We must consider the case that node_level is False but AcceleratorOnly is set. 
We should return None because AcceleratorOnly is not supported for the container level for now. 
Please add that case.

src/train/extractor/extractor.py
sunya-ch: Sorry. I miss that you also add mix of FeatureGroup.AcceleratorOnly in WorkloadOnly feature. 
I think we should modify function `get_workload_feature_data` to support all kind of features instead of separate the case here for the future extension. Could you please merge the logic to read `type` label in `get_workload_feature_data` function?

Yanbo0101: Do you mean to integrate the current logic of get_workload_feature_data into get_workload_feature_data? This requires additional returns.

Yanbo0101: Fixed All. Thanks! And according to the third point you mentioned, perhaps I don't need to deal with type labels anymore, and there won't be any additional return values.

Yanbo0101: Thanks for your reminder!  After some consideration, adding the logic to read type label and update features seems more appropriate.

sunya-ch: @Yanbo0101 
I am now pushing a PR https://github.com/sustainable-computing-io/kepler-model-server/pull/139 to export the trained models to the kepler-model-db which might have multiple conflict with the current code in the main brach. 
Please let me know if you want to have this PR merge first or need time to work on the feature extension.
This PR looks good to me if you are okay with single feature. I can merge and rebase on yours. 
Otherwise, you may need to work on rebase again.

Yanbo0101: I would like to have this PR merge first after your review. Thanks!

sunya-ch (approved):
/lgtm

PR #146, opened 2023-09-05
Update the feature normalization to use maxAbsScaler

marceloamaral:
The current code normalizes the feature data using [MinMaxScaler](https://github.com/scikit-learn/scikit-learn/blob/7f9bad99d/sklearn/preprocessing/_data.py#L277) or [StandardScaler](https://github.com/scikit-learn/scikit-learn/blob/7f9bad99d/sklearn/preprocessing/_data.py#L657).

The main problem with `MinMaxScaler` and `StandardScaler` is because they apply data [standardization](https://towardsdatascience.com/normalization-vs-standardization-quantitative-analysis-a91e8a79cebf) instead of simple normalization. In practices, standardization divides each feature by the standard deviation after the mean/min has been subtracted. The problem of subtracting values is because the resulted value could be negative, which will not make sense.

For example, let's consider the case of the MinMaxScaler, which calculates the following factors to transform the data.
```
feature_range = (0,1)
data_range = data_max - data_min
self.scale_ = (feature_range[1] - feature_range[0]) / data_range
self.min_ = feature_range[0] - data_min * self.scale_
```
Suppose we have CPU utilization at 110% and X.min at 100% (assuming the smallest workload, like stress-ng, consumes at least 100%). In a system with 10 CPUs, the maximum CPU utilization can reach 1000%. Thus, we have data_range = 100, self.scale_ = 0.01, and self.min_ = -1. The normalization (or transformation) is then applied as:
```
X *= self.scale_
X += self.min_
```
For CPU utilization of 110%, the normalized value becomes 0.1.

However, some workloads may result in CPU utilization lower than 100%, for example 20%. In this scenario, the normalized data would be calculated as `X = (X*self.scale_) + self.min_ = -0.8.`

Since in Kepler, we convert floating-point values to uint64, and uint64 cannot represent negative values, negative values overflow and return a huge positive value for power estimation.

In simple terms, the problem here is that the current scaling method can result in negative values for certain data, which leads to issues when converting these values to another data type (uint64), causing unexpected and impractical results in power estimation.

### DIFF
--- a/cmd/main.py
+++ b/cmd/main.py
@@ -652,7 +652,7 @@ def plot(args):
                     _ts_plot(power_data, power_cols, "Power source: {}".format(energy_source), output_folder, data_filename, ylabel="Power (W)")
     elif args.target_data == "estimate":
         from estimate import default_predicted_col_func
-        from sklearn.preprocessing import MinMaxScaler
+        from sklearn.preprocessing import MaxAbsScaler
 
         best_result_map, power_labels_map, best_model_id_map, _ = estimate(args)
         for energy_source, best_restult in best_result_map.items():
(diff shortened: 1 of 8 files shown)


sunya-ch (approved):
/lgtm 
Thank you for the update.

Issue #154, opened 2023-09-08
Reduce image size


sunya-ch:
### What would you like to be added?

We need to reduce image size.

```
quay.io/sustainable_computing_io/kepler_model_server_base                         latest                           2f1f98323e40   3 weeks ago     3.62GB
uay.io/sustainable_computing_io/kepler_model_server                              v0.6-test                        5b4c258f81eb   31 hours ago    3.65GB
quay.io/sustainable_computing_io/kepler_model_server                              v0.6                             8b0b737b3d16   6 days ago      3.62GB
```

We need to fix from the base image.

### Why is this needed?

Current kepler_model_server has 3.62GB (kepler is around 1.6GB) and might the root cause of the issue  https://github.com/sustainable-computing-io/kepler-model-server/issues/145.

### Suggestions
- We may remove keras and tensor flow for now as it is not used. 

After removing it, we got half of the space back
```
quay.io/sustainable_computing_io/kepler_model_server_base                         latest                           ede58a868812   15 seconds ago   1.65GB
```

Saurabhkr952:
I have reduced the size of the image using multi-stage builds. The original image size was 3.62 GB, but I was able to reduce it to 964 MB. I removed TensorFlow and Keras from the image, as they were not being used. I also used a smaller base image, which helped to reduce the size of the image.

Here is a screenshot of the image size: 
![image](https://github.com/sustainable-computing-io/kepler-model-server/assets/32189783/1a9d2ec2-ff13-44bb-8474-d20fdf7fd687)

sunya-ch:
> I have reduced the size of the image using multi-stage builds. The original image size was 3.62 GB, but I was able to reduce it to 964 MB. I removed TensorFlow and Keras from the image, as they were not being used. I also used a smaller base image, which helped to reduce the size of the image.
> 
> Here is a screenshot of the image size: ![image](https://user-images.githubusercontent.com/32189783/266829161-1a9d2ec2-ff13-44bb-8474-d20fdf7fd687.png)

@Saurabhkr952 
That's great. I also pushed the PR for removing the keras and tensor flow from the image too. 
https://github.com/sustainable-computing-io/kepler-model-server/pull/155
Could you please review the change on that PR and push another PR for adding the difference such as multi-stage build.

PR #155, opened 2023-09-08
remove keras dependency and separate base image

sunya-ch:
This PR is to temporarily remove keras until we have a proof of it need again to fix the issue https://github.com/sustainable-computing-io/kepler-model-server/issues/154.


marceloamaral (approved):
Makes sense to me.
/lgtm

rootfs (approved):

PR #168, opened 2023-09-19
Enhancement: Optimize Docker image size to reduce it below 1 GB (#154)

Saurabhkr952:
Pull Request: Reduce Image Size and Vulnerability

Changes:
- Utilize `python3.8-slim` as the base image to reduce image size and vulnerability.
- Added `--no-cache-dir` flag during `requirements.txt` installation for further size reduction.

To-Do:
- Remove [`Dockerfiles/Dockerfile.base`](https://github.com/sustainable-computing-io/kepler-model-server/tree/main/dockerfiles) as it's no longer required.

- Delete [`.github/workflows/docker-base-image.yml`](https://github.com/sustainable-computing-io/kepler-model-server/tree/main/.github/workflows) as `python3.8-slim` is used as the base image in the Dockerfile.

This pull request enhances image efficiency and security. Please review and merge as appropriate.

### DIFF
--- a/dockerfiles/Dockerfile.base
+++ b/dockerfiles/Dockerfile.base
@@ -1,4 +1,3 @@
-FROM registry.access.redhat.com/ubi8/ubi
-RUN  yum install -y python3.8 python3-devel gcc-c++  && yum clean all
-COPY requirements.txt requirements.txt
-RUN python3.8 -m pip install -r requirements.txt
+FROM python:3.8-slim
+COPY requirements.txt .
+RUN pip install --no-cache-dir -r requirements.txt


sunya-ch:
@Saurabhkr952  How long does it take for image to install all required modules? How's about just changing the Dockerfile.base to use python3.8-slim instead of ubi?

@rootfs What do you think about removing the base image?

rootfs:
@sunya-ch i haven't checked the image size yet but if the new image is significantly smaller and doesn't impact the model server functions, it then looks good to me.

sunya-ch:
- For build overhead, I think it depends on the container engine. 
It took me more than 200s for pip install with default docker setting.
```
=> [ 3/14] RUN pip install --no-cache-dir -r requirements.txt                                                                                                                                                                                                           218.7s
```
- I don't think there is a significant size change between using base image or not. The reduced size comes from changing from `ubi (229MB) + python install` to `python3.8-slim (154MB)`

```
python                                                                            3.8-slim                         b28341c2f98e   3 weeks ago      154MB
registry.access.redhat.com/ubi8/ubi                                               latest                           36d50e000e3b   3 weeks ago      229MB
```

- Unit test looks good without `python3-devel` and `gcc-c++`.  Not test integration yet.

To fix the error on test, you need to change the requirement copy line to

```
COPY ./dockerfiles/requirements.txt requirements.txt
```

Could you please try modifying that and amend the commit?
I would like to see the overhead on CI container runtime engine and test result. 

In my opinion, we should still use the Dockerfile.base but change to your suggested image and without devel and gcc install if all tests are passed.

sunya-ch (approved):
/lgtm 

[@Saurabhkr952](https://github.com/Saurabhkr952) Thank you for your contribution :)

PR #169, opened 2023-09-20
update model_training process (train and export)

sunya-ch:
** Rebase from the PR https://github.com/sustainable-computing-io/kepler-model-server/pull/172 **

Also, refer to archived pipeline that should be merged on kepler-model-db first by https://github.com/sustainable-computing-io/kepler-model-db/pull/16

Need 

- https://github.com/sustainable-computing-io/kepler-model-server/pull/172 
- https://github.com/sustainable-computing-io/kepler-model-db/pull/16

to be merged first.

---

This PR includes
- document (README) auto-generation on export script (auto-generate document mentioned in https://github.com/sustainable-computing-io/kepler-model-db/pull/14)
- sort metadata with feature group first
- handle missing trainer class in the list
- update get_url to also handle weight url
- add assure_path disable option for getting remote machine path
- add trainers_with_weight list to list model with weight
- update train script to use only stressng benchmark
- update stressng benchmark according to the result in community votes (19/09/2023)


model_training/script.sh
```diff
@@ -180,7 +181,7 @@ function quick_collect() {
 }
 
 function train() {
-    train_model stressng_kepler_query,coremark_kepler_query,parsec_kepler_query ${VERSION}
+    train_model stressng_kepler_query ${VERSION}
```
marceloamaral, line 184: Is this selecting the data only for training?

If it is also for the validation: should we also use the coremark results for testing to verify the accuracy of the model with different workload?

sunya-ch: It will be shuffled and use 10% of it for validation.
https://github.com/sustainable-computing-io/kepler-model-server/blob/6b4f716764dcf2ad7b1dc0371554c3acd8d7457b/src/train/trainer/__init__.py#L28

We need to refactor the code to have a fixed validation dataset. Should be created for a separate issue.

rootfs:
@sunya-ch the CI passed now.

rootfs (approved):

Issue #181, opened 2023-10-31
Support third party metrics in the model server


LeiZhou-97:
### What would you like to be added?

Support third party metrics in the model server

I go through the source code quickly, the model server seems only support to collect hard-core metrics for training.

What I want to know is if there are any development plans to support this feature in the near future? Thanks!

### Why is this needed?

End-user may have their own special metrics and these metrics are not included in Kepler.

sunya-ch:
[This function](https://github.com/sustainable-computing-io/kepler-model-server/blob/421ab2e553003c0eb5cde2ab45a67fa19ffbdbb7/cmd/main.py#L205) in command tool allows querying metric from prometheus with specific metric prefix.

However, I think we can modify the main src to support that too.
1. We can remove filter of kepler query in the following line to collect all the metric.
https://github.com/sustainable-computing-io/kepler-model-server/blob/421ab2e553003c0eb5cde2ab45a67fa19ffbdbb7/src/train/prom/prom_query.py#L33
The query result will be a mapping from query to the values in the same way we have with the kepler.

2. We need to provide an extension in `feature_to_query` in the same we do for node-level feature.

      https://github.com/sustainable-computing-io/kepler-model-server/blob/421ab2e553003c0eb5cde2ab45a67fa19ffbdbb7/src/train/extractor/extractor.py#L139
    
    
    https://github.com/sustainable-computing-io/kepler-model-server/blob/421ab2e553003c0eb5cde2ab45a67fa19ffbdbb7/src/util/prom_types.py#L45-L50

PR #185, opened 2023-11-06
Support third party metrics in model server

LeiZhou-97:
End user may have some specific metrics that are not included in kepler, so this patch enhance the model server to collect third party metrics

Limitation:
the pattern of thirdparty_metrics must align with kepler container level metrics, which contains containerid, containername and etc. Because it can reuse kepler container power consumption as output for training.
```
thridparty_metrics + other metrics(like cgroup....) -> input
container power consumption from kepler -> output
```


sunya-ch:
Thank you for your contribution. 
I believe it is related to the issue https://github.com/sustainable-computing-io/kepler-model-server/issues/181.
This is very promising to dynamically add third-party feature group. 
If all test is pass, this PR looks good to me. 

In my understanding, it will be set only when we set it in cmd argument. 
I would be nice if we can also call `update_thirdparty_metrics` when the third party metrics are set via environment or configmap for the model server, online trainer, and estimator sidecar. However, this could be done on the another PR if you don't want to implement it here.

LeiZhou-97:
@sunya-ch 
Thanks for your quick reply. I will update my code to fix the CI issue.

In my original design, I want to add the metrics into configmap and watch it in realtime. If it changes, the model server will update accordingly. This approach has some more effort and I have no much time. I can move it to the environment to align with existing configs as an work around.

sunya-ch:
> @sunya-ch Thanks for your quick reply. I will update my code to fix the CI issue.
> 
> In my original design, I want to add the metrics into configmap and watch it in realtime. If it changes, the model server will update accordingly. This approach has some more effort and I have no much time. I can move it to the environment to align with existing configs as an work around.

Thanks a lot. Nice work.

Actually, if you call the function `getConfig` in [util module ](https://github.com/sustainable-computing-io/kepler-model-server/blob/9f670d228beca0583a9dce8da4b47f731a08fc82/src/util/config.py#L43), it will try reading environment value or config data if set.  

You may define a key for the third-party metric such as `THIRD_PARTY_METRICS` then call getConfig('THIRD_PARTY_METRICS', "") to get the value. This key will be applied for both direct environment variable and config map data without any modification.

LeiZhou-97:
I've updated my code. Now it can get thirdparty_metrics from cmd and configmap/environment.

I also run the unittest on my server and it passed. It my need your approve to re-trigger the GHA.

PR #186, opened 2023-11-07
minor bug fixes to prepare a kind cluster

knarayan:
The command to prepare a kind cluster is failing.
```
./script.sh prepare_cluster
```
The directory `manifests` was supposed to be inside the directory `model_training/custom-cluster/kind` as per the script `model_training/custom-cluster/kind/kind.sh`. Hence moving it from `model_training/kind`.


knarayan:
In particular, the `cp` command [here](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/model_training/custom-cluster/kind/kind.sh#L75) in the `kind.sh` script is failing.

sunya-ch (approved):
/lgtm

Thank you for the fix!

PR #191, opened 2023-11-10
changes to collect metrics from Prometheus with benchmark run outside of kepler-model-server

knarayan:
Users might want to run a custom benchmark outside of `kepler-model-server` and collect metrics using `kepler-model-server` to train model subsequently.

#### Collect metrics (only query Prometheus to fetch metrics in the specified time window) 
    NATIVE="true" ./script.sh custom_collect
#### Validate the collected metrics
    NATIVE="true" ./script.sh validate customBenchmark
#### Train custom benchmark metrics
    NATIVE="true" ./script.sh custom_train

Note that Prometheus is queried to collect both `kepler` provided power consumption and `node-exporter` provided CPU utilization metrics (like `node_cpu_seconds_total`) for each cluster node.


sunya-ch:
Thank you for the PR. It is promising to extend the custom metrics for validation and train. 

I think we don't need to create a new function for them. We can add an option for specifying `start-time`,`end-time`  directly in addition to `interval`. In stead of saving only query result (like idle.json), we have custom_benchmark.json and custom_query_response.json when `start_time` `end_time` are specified with `startTimeUTC`/`startTimeUTC` in json as you defined. 
For the custom_metrics, is it the same as PROM_THIRDPARTY_METRICS introduced in https://github.com/sustainable-computing-io/kepler-model-server/pull/185?

src/train/exporter/validator.py
```diff
@@ -7,7 +7,7 @@
 from loader import load_train_args
 from config import ERROR_KEY
 
-required_benchmark = ["stressng_kepler_query"]
+required_benchmark = ["sample_kepler_query", "stressng_kepler_query", "customBenchmark_kepler_query"]
```
sunya-ch, line 10: required benchmark is defined for making sure that the required benchmark is included when exporting the model. sample_kepler_query and customBenchmark_kepler_query should not be included here as it is optional.

src/util/prom_types.py
```diff
@@ -18,6 +18,7 @@
 PROM_THIRDPARTY_METRICS = getConfig('PROM_THIRDPARTY_METRICS', "").split(',')
 
 metric_prefix = "kepler_"
+custom_metrics = "node_cpu_seconds_total"
```
sunya-ch, line 21: If this is fixed, it should be defined in feature group.
If it is custom definition, it should be able to set dynamically by design of third-party metric introduced by this PR https://github.com/sustainable-computing-io/kepler-model-server/pull/185.

model_training/README.md
```diff
@@ -106,20 +136,21 @@ Run
 
 1. Fork `kepler-model-db`.
 
-1. Validate and make a copy by export command. Need to define `machine id`, `local path to forked kepler-model-db/models`, and `author github account`. 
+1. Validate and make a copy by export command. Need to define `machine id`, `local path to forked kepler-model-db/models`, `author github account` and `benchmark type`.
 
     Run
     ```
-    ./script.sh export <machine id> <path to kepler-model-db/models> <author github account>
+    ./script.sh export <machine id> <path to kepler-model-db/models> <author github account> <benchmark type>
```
sunya-ch, line 143: I would like to avoid fixing the benchmark type here since we also expect the mix type of benchmarks (both standard and custom).

model_training/script.sh
```diff
@@ -176,23 +182,35 @@ function collect() {
...
+    custom_collect_data customBenchmark
+}
+
 function quick_collect() {
     collect_data sample cpe-operator-system 10
 }
 
 function train() {
-    train_model stressng_kepler_query ${VERSION}
+    train_model stressng_kepler_query ${VERSION}_stressng
```
sunya-ch, line 194: I would like to keep only passing $VERSION here.
To separate custom pipeline, `PIPELINE_PREFIX` should be set.

model_training/README.md
sunya-ch, line 143: I would like to avoid setting benchmark type in export command since we also expect mix of CPE-based benchmark and custom benchmark.

model_training/README.md
```diff
@@ -106,20 +136,21 @@ Run
...
     ```
 
-    If you are agree to also share the raw data (preprocessed data and archived file of full pipeline), run
+    If you also agree to share the raw data (preprocessed data and archived file of full pipeline), run
 
     ```
-    ./script.sh export_with_raw <machine id> <path to kepler-model-db/models> <author github account>
+    ./script.sh export_with_raw <machine id> <path to kepler-model-db/models> <author github account> <benchmark type>
```
sunya-ch, line 149: same concern as above about mix-type benchmark set.

sunya-ch (commented):
Thank you so much for the PR. I think this PR is very helpful on introducing a new custom benchmark which is not based on CPE operator. Without CPE-defined benchmark, trainer can directly set start time and end time (or last interval) of the benchmark. 

However, since it has some conflict on the current design and recently-merged PR (https://github.com/sustainable-computing-io/kepler-model-server/pull/185), I would like to have some more discussion on changes.

After we got the conclusion, it would be great if contributor would also update the instruction in https://github.com/sustainable-computing-io/kepler-model-server/blob/main/contributing.md#introduce-new-benchmarks.

src/util/prom_types.py
knarayan: I agree @sunya-ch , there is no need of `custom_metrics` since third-party metric helps achieve the same.

model_training/script.sh
knarayan: I have set this `suffix` similar to the way it is set for the sample benchmark [here](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/model_training/script.sh#L188).
Thus, in the source code, if the benchmark == `sample`, then it is set to `_sample`, if the benchmark == `stressng`, then it is set to `_stressng`, and if the benchmark == `customBenchmark`, then it is set to `_customBenchmark`. Also, I assume that `PIPELINE_PREFIX` can be used to set the appropriate prefix (where as we are setting here a suffix).
I hope this addresses your comment.

src/train/exporter/validator.py
knarayan: The setting [`MAIN_COLLECT_INPUT="stressng"`](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/model_training/script.sh#L214) along with [`required_benchmark = ["stressng_kepler_query"]`](https://github.com/sustainable-computing-io/kepler-model-server/blob/main/src/train/exporter/validator.py#L10) only lets exporting models trained using the benchmark `stressng`. Ideally, it should allow any other benchmark (like `sample`, `customBenchwork` or some other future workload added to the collection of benchmarks).

src/util/prom_types.py
```diff
@@ -7,13 +7,17 @@
 PROM_HEADERS = ''
 PROM_QUERY_INTERVAL = 300
 PROM_QUERY_STEP = 3
+PROM_QUERY_START_TIME = '2023-11-13T02:47:57Z'
```
sunya-ch, line 10: should not be static. We might define it based on the running time like below or set to '' to be handled if empty later. 

        PROM_QUERY_END_TIME = datetime.datetime.now()
        PROM_QUERY_START_TIME = PROM_QUERY_END_TIME - datetime.timedelta(seconds=PROM_QUERY_INTERVAL)

model_training/script.sh
sunya-ch: Got it. This also makes sense to me.

src/train/exporter/validator.py
sunya-ch: The original design is to avoid the overfitting benchmark suite. 
So, the concept of required benchmark here is to assure that stressng should be included in the benchmark suite. 
However, if we will allow any benchmark suite (no need to include stressng), we can just remove this required benchmark check.  We can add the checklist to have stressng workload manually on PR pushed to the kepler-model-db repo. 
Would you mind noting the issue to remove this requirement check on the model server source side?

src/util/prom_types.py
knarayan: Ok @sunya-ch , I have set it to empty. Hope this is also a valid setting.

src/train/exporter/validator.py
knarayan: Sure @sunya-ch , did this update.

sunya-ch (approved):
@knarayan Thank you so much for your contribution. 

As there are multiple significant changes here, I would like to summarize the contribution of this PR again as below.
This PR includes
- introducing customBenchmark with variable-specified startTime, endTime for non-CPE benchmark (separate sample, stressng, and customBenchmark)
- supporting validation for all available queries 
- removing required benchmark constraint on export function to allow reuse on customBenchmark

Please feel free to add or correct my summary.

PR #199, opened 2023-12-01
[CI] image build logic enhancement

SamYuan1990:
- use latest GHA to manage container image build instead of make script.
- update logic between base image and image itself
- add arm64 and s390 image support


SamYuan1990:
> @SamYuan1990 The changes mostly look good to me. Is it possible to have the base image rebuild only when the relevant files are changed (Dockerfile.base and requirements.txt)? https://how.wtf/run-workflow-step-or-job-based-on-file-changes-github-actions.html

I will try, there two items.
1. if the base image updates, we need to build from base image.
2. if there no changes on base image, we need to build model server image any way.

but, I am curious that it seems before this PR. if we update (for example, requirement.txt) and I didn't find a line in code limit the model server image build after base image?

SamYuan1990:
`ERROR: failed to solve: quay.io/sustainable_computing_io/kepler_model_server_base:v0.7: pulling from host quay.io failed with status code [manifests v0.7]: 502 Bad Gateway` @sunya-ch could you please help rerun the test case? https://github.com/sustainable-computing-io/kepler-model-server/actions/runs/7086391503/job/19284563239?pr=199

SamYuan1990:
@sunya-ch , could you please help review this PR?

sunya-ch (approved):
/lgtm Thank you.

.github/workflows/build-push.yml
```diff
@@ -2,42 +2,84 @@ name: BuildPushDeployImage
...
+    branches: [ main ]
 
 jobs:
-  build:
+
+  changes:
     runs-on: ubuntu-latest
+    outputs:
+      src: ${{ steps.changes.outputs.src }}
+      infra: ${{ steps.changes.outputs.infra }}
```
sunya-ch, line 13: I don't see infra map below. 
I think we need to remove it.

sunya-ch (changes requested):
I don't think we need `infra` in outputs. Other parts are good to go.

.github/workflows/build-push.yml
SamYuan1990: updated.

sunya-ch (approved):
Thank you!

Issue #187, opened 2023-11-07
upgrade xgboost to 2.0.1


rootfs:
### What would you like to be added?

current xgboost used in model server is 1.6.2 (https://github.com/sustainable-computing-io/kepler-model-server/blob/main/dockerfiles/requirements.txt#L10)

### Why is this needed?

please upgrade to 2.0.1 so kepler can sync with the [same version](https://github.com/sustainable-computing-io/kepler/pull/1007/files#diff-916182360fb71d1a73e352508cabb091f8ef85d2e0ade76ff8f3ee325d920154R26) and use the model in the same way.

Note, xgboost 2.0.1 may need python 3.8 and higher

PR #210, opened 2023-12-22
upgrade xgboost to 2.0.1

sunya-ch:
According to https://github.com/sustainable-computing-io/kepler-model-server/issues/187, this PR made the version upgrade of xgboost and update unit test workflow to build test image without base image if there is a change on base image related files (such as requirement.txt).

### DIFF
--- a/.github/workflows/unittest.yaml
+++ b/.github/workflows/unittest.yaml
@@ -3,14 +3,34 @@ on:
   pull_request:
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      src: ${{ steps.changes.outputs.src }}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dorny/paths-filter@v2
+      id: changes
+      with:
+        filters: |
+          src:
+            - 'dockerfiles/requirements.txt'
+            - 'dockerfiles/Dockerfile.base'
+            - '.github/workflows/unittest.yaml'
+
   run-test-cases:
     runs-on: ubuntu-latest
+    needs: changes
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/unittest.yaml
```diff
@@ -3,14 +3,34 @@ on:
...
+      src: ${{ steps.changes.outputs.src }}
+    steps:
+    - uses: actions/checkout@v4
+    - uses: dorny/paths-filter@v2
+      id: changes
+      with:
+        filters: |
+          src:
+            - 'dockerfiles/requirements.txt'
+            - 'dockerfiles/Dockerfile.base'
```
SamYuan1990, line 18: do we need to worried about any our dependency updates breaks CI? If I am get the points for line 17 and 18, which means any updates for dependency as xgboost version will skip unit test?
hence if any breaking changes happened in xgboost api, it will skipped CI and merged?

.github/workflows/unittest.yaml
```diff
@@ -3,14 +3,34 @@ on:
...
+            - '.github/workflows/build-push.yml'
```
SamYuan1990, line 19: why do we including build-push github action here?

.github/workflows/unittest.yaml
SamYuan1990: I got the code, cool~

.github/workflows/unittest.yaml
sunya-ch: Thank you @SamYuan1990  It should be the unittest not the build-push. Updated.

SamYuan1990 (approved):
LGTM

rootfs (approved):

Issue #220, opened 2024-01-24
[CI] refactor CI workflow 


sunya-ch:
### What would you like to be added?

Refactor CI workflow
### workflow on push
   - unit test
   - build image (push if docker-secret exists), build tag by commit,  when branch = main, build tag = v[version]
   - collect data (if aws-secret exists) - https://github.com/sustainable-computing-io/kepler-model-server/issues/219
   - train model (if aws-secret  exists)

####  When IMAGE_REPO, DOCKER_USERNAME, DOCKER_PASSWORD are not set

<img width="1416" alt="Screenshot 2024-01-25 at 16 38 11" src="https://github.com/sustainable-computing-io/kepler-model-server/assets/11749848/d2e9bcd7-5ce6-461a-a0ac-8fa9a07ed085">

#### When IMAGE_REPO, DOCKER_USERNAME, DOCKER_PASSWORD are set
<img width="1414" alt="Screenshot 2024-01-25 at 17 01 18" src="https://github.com/sustainable-computing-io/kepler-model-server/assets/11749848/72b7d80f-0e50-440b-9c5f-2538572d0f77">

### workflow on PR
   - unit test
   - build image (push if docker-secret exists), build tag by PR number
   - tekton test on kind with collect and train (VM)
    
    
- integration-test (as-is)
    

### Why is this needed?

Need to reduce redundancy and apply built model to collect and train data. The test will also remove dependency of unittest and integration test from the kepler-model-db.

Issue #205, opened 2023-12-04
Remove kubelet feature


sunya-ch:
Corresponding to change by https://github.com/sustainable-computing-io/kepler/pull/715

PR #222, opened 2024-01-25
add local db, remove kubelet, update push-pr

sunya-ch:
This PR refactors CI workflows to remove dependency to models uploaded to kepler-model-db as mentioned in https://github.com/sustainable-computing-io/kepler-model-server/issues/220.

Instead of using URL from kepler-model-db GitHub repository, this change instead uses results from pipeline run by the pushed commit by serving local model-db (http-server).

To reduce pipeline preparation time on testing, this change also update pipeline_test.py to support `minimal trainer`. 

local-model-db

In addition, this PR also removes kubelet feature from the code according to https://github.com/sustainable-computing-io/kepler-model-server/issues/205 and fixes some test assertion and handing path not exists.

Screenshot 2024-01-27 at 8 51 22

Note (tasks to done on different PRs):
- [ ] need to update document about availability of default models (kepler-model-db URL) 
- [x] need to correct Kepler to request weight with `intel_rapl` not `rapl`: https://github.com/sustainable-computing-io/kepler/pull/1204
- [ ] need to improve Kepler to support other simple models (curvefit) and request trainer corresponding to the call function
- [ ] need to separate machine ID for different node type (will push PR soon)


sunya-ch:
Note: on PR the base-image is skipped since secret is not available on PR and integration and tekton test requires new base image pushed. Can we add secrets to the PR to allow it to push base-image with tag pr-<pr number>? @rootfs

rootfs (approved):

Issue #216, opened 2024-01-17
Revisit model selection logic and results


sunya-ch:
### What would you like to be added?

Add similarity function to the model selection logic in addition to specific node type specify. 
In addition, include this similarity value on the selection to the model response.

### Why is this needed?

Currently, estimator sidecar or kepler local estimator sends a ModelRequest  to the server.
![Screenshot 2024-01-17 at 16 16 46](https://github.com/sustainable-computing-io/kepler-model-server/assets/11749848/b3afdf5d-e7d8-413f-b7b6-55dd0e49011e)

The selection logic includes only (1) checking available features, (2) equal comparison on model metadata, and (3) accuracy threshold. 
The reply contains  the model file (as a model zip or weight).

We need to report more about the selected model including the indicator to the model accuracy and the relation to the requesting machine which we are planning to report about similarity and reliability of the model.

sunya-ch:
closed by #370

PR #225, opened 2024-01-29
add node type indexing

sunya-ch:
This PR is related to the issue https://github.com/sustainable-computing-io/kepler-model-server/issues/216 and TODO list in previous PR https://github.com/sustainable-computing-io/kepler-model-server/pull/222.

The main change is to add node_type indexing in `src/train/profiler/node_type_index.py`. 
I introduce NodeTypeSpec, NodeTypeIndexCollection to group inputting machine data by spec on each pipeline training. 
As shown below, at collection, we will autogenerate NodeTypeSpec (processor, #cores, #chips, memory and so on) and keep it in data path. At training, we will read that value by the machine id and then index it in the NodeTypeIndexCollection. 
If the same spec has been indexed, it will use the same index number. However, we expect a step to append data from the same group before training. For AWS instance, we expect single profile per one index. The machine index will be kept under pipeline folder in Json format (node_type_index.json). We can read this file and generate machine index on export.

node_type_index

In addition to above enhancement, this PR also includes multiple bug fixes on CI workflow including adding complete-train pipeline run on tekton test.

### DIFF
--- a/.github/workflows/collect-data-self-hosted.yml
+++ b/.github/workflows/collect-data-self-hosted.yml
@@ -147,7 +147,7 @@ jobs:
         working-directory: model_training/tekton
         run: |
           kubectl apply -f tasks
-          kubectl apply -f tasks/s3-pusher
+          kubectl apply -f tasks/s3
           kubectl apply -f pipelines
       - name: Run Tekton Pipeline with S3Push
         run: |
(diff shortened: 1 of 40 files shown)


sunya-ch:
I will update exporter for separating each node type. Here are examples of exported value for the pipeline trained on SPECPower data. 

https://github.com/sunya-ch/kepler-model-db/blob/specpower/models/v0.7/README.md

rootfs (approved):

rootfs:
also cc @KaiyiLiu1234

PR #230, opened 2024-02-12
feat: add uml diagrams

vimalk78:
Generated class diagrams from source for better understanding of kepler-model-server


vimalk78:
Cc: @sunya-ch @rootfs

umls/README.md
sthaha, line 1: Could we move this under `docs/developer/`

vimalk78:
> This would be very helpful for developer. Thank you. Could you please move the directory as @sthaha suggested? After that, I think this PR is all good to merge.

@sunya-ch  moved

sunya-ch (approved):
Thank you!

PR #256, opened 2024-04-27
[CI nit] use a fixed kepler-action version as found some PR may change kubecon…

SamYuan1990:
use a fixed kepler-action version as found some PR may change kubeconfig path

kepler-action has it daily check to avoid it failure, same daily check is adding to local-dev-cluster.


sunya-ch (approved):
/lgtm

Issue #259, opened 2024-06-06
Use production WSGI Server instead of Development server 


sthaha:
### What happened?

Running model-server latest image produces the following log

```
model-server-1  |  * Serving Flask app 'model_server' (lazy loading)
model-server-1  |  * Environment: production
model-server-1  |    WARNING: This is a development server. Do not use it in a production deployment.
model-server-1  |    Use a production WSGI server instead.
model-server-1  |  * Debug mode: off
model-server-1  | WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
model-server-1  |  * Running on all addresses (0.0.0.0)
```

### What did you expect to happen?

Shouldn't show the warning above

### How can we reproduce it (as minimally and precisely as possible)?

Use kepler vm docker compose to run model-server 

### Kepler image tag

NA

### Deployment

- [X] Model server
- [ ] Estimator
- [ ] Online trainer
- [ ] Offline trainer
- [ ] Profiler

### Kepler model server image tag if deployed

<details>
0.7.7
</details>

### Kepler estimator image tag if deployed

<details>

</details>

### Kepler online trainer image tag if deployed

<details>

</details>

### Kepler offline trainer image tag if deployed

<details>

</details>

### Kepler profiler image tag if deployed

<details>

</details>

### Kubernetes version

NA

### Install tools

NA

### Kepler deployment config

NA

PR #263, opened 2024-06-13
WIP Include Production WSGI (Gunicorn) to replace Flask Default Server

KaiyiLiu1234:
To Address: https://github.com/sustainable-computing-io/kepler-model-server/issues/259

A side effect of using Gunicorn is that it will replace the address with default address and default port (8000). This can be resolved by binding (ex. binding model_server to 0.0.0.0:8100). Binding with 0.0.0.0 is acceptable but according to [Flask](https://flask.palletsprojects.com/en/3.0.x/deploying/gunicorn/), it is better to also introduce nginx server to act as a reverse proxy. Will this be necessary to include? 

Also, since model_server AND offline_trainer (and in future online_trainer) expects to use their own Flask Apps, supervisord is needed to run multiple CMD for model_server and offline_trainer in Dockerfile. 

Currently, this changes here works for metal docker compose so long as  python3.8 -u src/server/model_server.py is replaced by gunicorn -b -0.0.0.0.8100 -src.server.model_server:app. However, there is more functionality that can be introduced depending on what is acceptable or not.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -18,4 +18,6 @@ EXPOSE 8101
 # port for Offline Trainer
 EXPOSE 8102
 
+CMD ["gunicorn", "--bind", "0.0.0.0:8100", "src.server.model_server:app"]
+
 ENTRYPOINT ["python3.8", "cmd/main.py"]
--- a/dockerfiles/Dockerfile.test
+++ b/dockerfiles/Dockerfile.test
@@ -24,4 +24,5 @@ EXPOSE 8101
 # port for Offline Trainer
 EXPOSE 8102
 
-CMD [ "python3.8", "-u", "src/server/model_server.py" ]
+#CMD [ "python3.8", "-u", "src/server/model_server.py" ]
+CMD ["gunicorn", "--bind", "0.0.0.0:8100", "src.server.model_server:app"]
--- a/dockerfiles/Dockerfile.test-nobase
+++ b/dockerfiles/Dockerfile.test-nobase
@@ -25,4 +25,5 @@ EXPOSE 8101
 # port for Offline Trainer
 EXPOSE 8102
 
-CMD [ "python3.8", "-u", "src/server/model_server.py" ]
+# CMD [ "python3.8", "-u", "src/server/model_server.py" ]
+CMD ["gunicorn", "-b", "0.0.0.0:8100", "src.server.model_server:app"]
--- a/dockerfiles/requirements.txt
+++ b/dockerfiles/requirements.txt
@@ -1,4 +1,5 @@
 flask==2.1.2
+gunicorn==22.0.0
 Werkzeug==2.2.2
 protobuf==3.19.4
 pandas==1.4.4


sthaha:
@KaiyiLiu1234 https://developers.redhat.com/articles/2023/08/17/how-deploy-flask-application-python-gunicorn#containerization is good reference to follow

> Binding with 0.0.0.0 is acceptable but according to [Flask](https://flask.palletsprojects.com/en/3.0.x/deploying/gunicorn/), it is better to also introduce nginx server to act as a reverse proxy. Will this be necessary to include?

I am not sure if the recommendation applies to containers / k8s. 

> Gunicorn should not be run as root because it would cause your application code to run as root, which is not secure. However, this means it will not be possible to bind to port 80 or 443. Instead, a reverse proxy such as [nginx](https://flask.palletsprojects.com/en/3.0.x/deploying/nginx/) or [Apache httpd](https://flask.palletsprojects.com/en/3.0.x/deploying/apache-httpd/) should be used in front of Gunicorn.

The part applies to containers  is don't run as root. However, binding to port 80/443 isn't needed for containers. 

> You can bind to all external IPs on a non-privileged port using the -b 0.0.0.0 option. Don’t do this when using a reverse proxy setup, otherwise it will be possible to bypass the proxy.

Again, this is fine for a containerized env.  I feel following the article linked should be good enough to fix our issue.

KaiyiLiu1234:
@sthaha @sunya-ch I incorporated the Gunicorn changes into a test dockerfile for now. How we interact with the image will change now in docker composes and kubernetes resources (for instance command should not be python model_server.py - it should instead use gunicorn or no command as the CMD for the dockerfile can handle launching the gunicorn servers for model server, offline trainer and online trainer). I can also include in the makefile commands to run the new gunicorn test dockerfile in this PR or in a future PR?

Makefile
```diff
@@ -51,27 +55,42 @@ run-model-server:
 	$(CTR_CMD) run -d --platform linux/amd64 -e "MODEL_TOPURL=http://localhost:8110" -v ${MODEL_PATH}:/mnt/models -p 8100:8100 --name model-server $(TEST_IMAGE) /bin/bash -c "python3.8 tests/http_server.py & sleep 10 &&  python3.8 src/server/model_server.py"
 	while ! docker logs model-server | grep -q Serving; do   echo "waiting for model-server to serve";  sleep 5; done
 
+run-model-server-gunicorn-complete:
```
sthaha, line 58:
```suggestion
run-model-server-prod:
```

Makefile
```diff
@@ -51,27 +55,42 @@ run-model-server:
...
+	$(CTR_CMD) run -d --platform linux/amd64 -e "MODEL_TOPURL=http://localhost:8110" -v ${MODEL_PATH}:/mnt/models -p 8105:8105 -p 9109:9109 --name model-server-gunicorn-complete $(GUNICORN_TEST_IMAGE)
```
sthaha, line 59:
```suggestion
	$(CTR_CMD) run -d \
	  --platform linux/amd64 \
	  -e "MODEL_TOPURL=http://localhost:8110" \
	  -v ${MODEL_PATH}:/mnt/models \
	   -p 8105:8105 -p 9109:9109 \
	   --name model-server-prod \
	    $(GUNICORN_TEST_IMAGE)
```

cmd/base_gunicorn_config.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+
+workers = int(os.environ.get('GUNICORN_PROCESSES', '2'))
+
+threads = int(os.environ.get('GUNICORN_THREADS', '4'))
+
+port = os.environ.get('GUNICORN_PORT', '8100')
+
+bind = os.environ.get('GUNICORN_BIND', '0.0.0.0:' + port)
```
sthaha, line 10:
```suggestion
workers = int(os.environ.get('GUNICORN_PROCESSES', '2'))
threads = int(os.environ.get('GUNICORN_THREADS', '4'))
port = os.environ.get('GUNICORN_PORT', '8100')
bind = os.environ.get('GUNICORN_BIND', '0.0.0.0:' + port)
```

cmd/base_gunicorn_config.py
sthaha, line 1: how about we call it gunicorn_config.py ?

sthaha: lets keep config under a `config` dir

dockerfiles/Dockerfile.test-gunicorn
```diff
@@ -0,0 +1,34 @@
...
+
+# port for Model Server
+EXPOSE ${PORT_MODEL_SERVER}
+# port for Online Trainer (TODO: reserved for event-based online training)
+EXPOSE ${PORT_ONLINE_TRAINER}
+# port for Offline Trainer
+EXPOSE ${PORT_OFFLINE_TRAINER}
+
+# ENTRYPOINT ["python3.8", "cmd/main.py"]
+#ENTRYPOINT [ "python3.8", "-u", "src/server/model_server.py" ]
```
sthaha, line 33:
```suggestion
```

dockerfiles/Dockerfile.test-gunicorn
```diff
@@ -0,0 +1,34 @@
...
+RUN mkdir -p /usr/local/resource
+
+COPY dockerfiles/requirements.txt dockerfiles/requirements.txt
+RUN pip install --no-cache-dir -r dockerfiles/requirements.txt
+
+COPY src/estimate src/estimate
+COPY src/server src/server
+COPY src/train src/train
+COPY src/util src/util
+COPY cmd cmd
```
sthaha, line 15: why not copy the entire src and cmd ?

dockerfiles/Dockerfile
```diff
@@ -18,4 +18,5 @@ EXPOSE 8101
 # port for Offline Trainer
 EXPOSE 8102
 
+
```
sthaha, line 21: unrelated change?

dockerfiles/Dockerfile.test-gunicorn
sunya-ch: I think it is my bad that when we run the test the built ./model will be inside the src folder. We need to fix that first then the src will be cleaned enough for entire copy.

KaiyiLiu1234: I can check it out.

cmd/gunicorn.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+echo "Starting Model Server"
+export GUNICORN_PORT=$PORT_MODEL_SERVER
+gunicorn --config config/gunicorn_config.py src.server.model_server:app &
+echo "Starting Offline Trainer"
+export GUNICORN_PORT=$PORT_OFFLINE_TRAINER
+gunicorn --config config/gunicorn_config.py src.train.offline_trainer:app &
+wait
```
sthaha, line 9: Can we create separate containers for model-server and trainer? 
This removes the issue of handling (unexpected) termination.

dockerfiles/Dockerfile.test-gunicorn
```diff
@@ -0,0 +1,33 @@
...
+ENV PORT_OFFLINE_TRAINER=9105
+
+# port for Model Server
+EXPOSE ${PORT_MODEL_SERVER}
+# port for Online Trainer (TODO: reserved for event-based online training)
+EXPOSE ${PORT_ONLINE_TRAINER}
+# port for Offline Trainer
+EXPOSE ${PORT_OFFLINE_TRAINER}
+
+CMD ["sh", "cmd/gunicorn.sh"]
```
sthaha, line 33: Is this idea to replace current `Dockerfile` with this?

sunya-ch:
@KaiyiLiu1234 Can we close this PR?

KaiyiLiu1234:
@sunya-ch Let's do it. We can reincorporate this in the future.

PR #272, opened 2024-07-03
chore(python): upgrade to 3.10

sthaha:
NOTE: - https://devguide.python.org/versions/ - 3.8 will soon be EOL
This buys us some more time upgrading to 3.11 or 3.12 which fails (`hatch shell`) at the moment and requires further investigation.


sunya-ch (approved):
/lgtm thanks!

PR #274, opened 2024-07-03
chore(pyproject): fix pip install not installing all packages

sthaha:
This fixes `pip install .` which installs `model-server` and `estimator` that 
also requires `util` and `train` to be installed as well.


vprashar2929 (approved):

sunya-ch (approved):
/lgtm

PR #276, opened 2024-07-04
ci: do not overwrite already published images

sthaha:


.github/workflows/push-to-main.yml
```diff
@@ -84,4 +84,4 @@ jobs:
       self_hosted_github_token: ${{ secrets.GH_SELF_HOSTED_RUNNER_TOKEN }}
       aws_access_key_id: ${{ secrets.AWS_ACCESS_KEY_ID }}
       aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-      aws_region: ${{ secrets.AWS_REGION }}
```
sunya-ch, line 87: please add back the line at the end of file. then this PR should be good to go.

sthaha: Actually, this adds a missing newline instead of removing it.

sunya-ch (approved):

.github/workflows/push-to-main.yml
sunya-ch: Oh sorry. Then it should good to go.

PR #283, opened 2024-07-06
Revert to latest tag instead of v0.7

KaiyiLiu1234:
Reverting to latest tag resolves python3.10 missing error.


model_training/custom-cluster/kind/local-registry.yml
```diff
@@ -0,0 +1,10 @@
+---
```
sunya-ch, line 1: Is this config file in use by this repo?

sunya-ch (approved):
Tekton image version change looks good to me.

sthaha (changes requested):
Looks good overall, but requesting change only to ensure that  if  `[cluster/kind/local-registry.yml](https://github.com/sustainable-computing-io/kepler-model-server/pull/283/files#diff-ec25f9e92cb139bda90bebf5f93d9c3e39cf2a13ba4985010ad7ef32117c2b51)` is not checked in by mistake.

sthaha (approved):

PR #284, opened 2024-07-08
fix(model-training): revert to python3.8 for s3 scripts

sthaha:
The[ image used for s3](https://quay.io/repository/sustainability/kepler_model_server/s3/manifest/sha256:c92c40a0e2974146d8d87931bc5d36e062105e23b3660ec0cb40183dc979c63f) step still uses python 3.8 and needs to be rebuilt before switching to python 3.10


sthaha:
Closing this PR to use 3.10 image instead.

PR #286, opened 2024-07-08
chore: replace requirements.txt with pyproject

sthaha:


sunya-ch (approved):
This is great. Thank you!

PR #287, opened 2024-07-08
chore(s3): use pyproject.yaml for s3 scripts

sthaha:


sunya-ch (approved):
Look great to me. Thanks! -> Sorry but could you check the change request below.

sunya-ch:
Note that, we need to push the s3 image on push-to-main.yml and also need to update default image to the model_training/tekton/tasks/s3/*.yaml to use that pushed image.

model_training/tekton/tasks/s3/ibmcloud-s3-load.yaml
```diff
@@ -45,9 +45,8 @@ spec:
           secretKeyRef:
```
sunya-ch, line 45: need to change image to use kepler_model_server_s3 ?
same as other s3/*.yaml

sunya-ch (changes requested):

.github/workflows/build-push.yml
```diff
@@ -82,22 +88,31 @@ jobs:
...
           registry: ${{ inputs.image_repo }}
           username: ${{ secrets.docker_username }}
-          password: ${{ secrets.docker_password }}    
+          password: ${{ secrets.docker_password }}
+    - name: Build-push s3 image
+      if: ${{ needs.check-secret.outputs.available == 'true' && env.s3_change == 'true' }}
+      uses: docker/build-push-action@v6
+      with:
+        platforms: linux/amd64
+        push: true
```
sunya-ch, line 102: How's about having 
`push: ${{ inputs.push }}`
This is unlike base image. We don't have to push it for all cases. We should treat this the same as the main image.

model_training/tekton/tasks/s3/aws-s3-load.yaml
```diff
@@ -23,7 +23,7 @@ spec:
     optional: true
   steps:
     - name: load
-      image: quay.io/sustainability/kepler_model_server/s3:v0.7-py3.10
+      image: quay.io/sustainable_computing_io/kepler_model_server/s3:latest
```
sunya-ch, line 26: I think we should add new input for the s3 image. 
For the PR, we can use a temporarily built image for test. applied this comment to other yaml.

sunya-ch (changes requested):

.github/workflows/build-push.yml
sthaha: In which case do we not need to push? Currently it gets built-and-pushed when s3 changes.

sunya-ch: It is the same reason that we didn't push the PR image to the repo. If we decide to keep all PR on the repo, we should also do the same to the main image. 

Another issue I request for change is that we want to have the same s3 image tag tested in the tekton and integration test but currently it refers to the latest tag. The s3 built here is with PR tag. You have to update the tekton/tasks/s3/*.yaml to use the same image tag.

sthaha: Since build and push is run only after merge, and as the training pipeline refers to latest, the training pipeline should make use of the latest image.

sunya-ch: You are correct. I miss the point that we cannot test s3 on PR anyway.

sunya-ch (approved):
/lgtm Thanks a lot.

PR #293, opened 2024-07-08
ci: build base image if a change is detected

sthaha:
This PR builds (but not push) base image if a change is detected in critical files.

This should help detect issues such as https://github.com/sustainable-computing-io/kepler-model-server/actions/runs/9834078233 where the build-and-push base failed after the PR was merged.


sunya-ch (commented):
/lgtm

With some note that, on PR, the kepler-model-server will not be tested with the built base image. We need another PR to build the base image to temporary tag and use that to build the kepler-model-server image for testing.

sunya-ch (approved):

PR #296, opened 2024-07-09
chore(gha): fix s3 changed flag

sthaha:


sunya-ch (approved):
good to go! Thanks for the fix.

PR #298, opened 2024-07-09
fix(tekton): update the README

vprashar2929:
This commit mainly fixes the secret example snippet according to the cloud provider, and also fences the code blocks along with removing the necessary whitespaces


vprashar2929:
cc: @sthaha

sthaha:
@vprashar2929 Thank you for the patch. 
Could you please use `doc:` instead of `fix:` if you are using conventional commits? 
See: https://github.com/qoomon/git-conventional-commits

sthaha (approved):

PR #319, opened 2024-07-22
chore(pyproject): upgrade deps

sthaha:
This replaces all dependabot upgrades  with a single update that  resolves 
conflict when upgrading individual package.

See: https://github.com/sustainable-computing-io/kepler-model-server/pulls/app%2Fdependabot


tests/pipeline_test.py
```diff
@@ -1,11 +1,6 @@
-import os
-import sys
-
-#################################################################
-# import internal src 
-src_path = os.path.join(os.path.dirname(__file__), '..', 'src')
-sys.path.append(src_path)
```
sunya-ch, line 7: Removing the path append here causes the failed on test. Is there a reason to remove this?

PR #321, opened 2024-07-24
chore(pyproject): upgrade dependencies

sthaha:
Code changes required are mostly to address the exception raised when
computing mean of columns where some of the columns contained non
numeric data.

E.g. df.mean()[numeric_cols] is now changed to df[numeric_cols].mean()

---

This replaces all dependabot upgrades with a single update that resolves
conflict when upgrading individual package.

See: https://github.com/sustainable-computing-io/kepler-model-server/pulls/app%2Fdependabot


sunya-ch (approved):
LGTM! Thanks.

PR #323, opened 2024-07-24
chore: Use top-level kepler_model python package

sthaha:
The PR refactors the code base to follow python packaging guidelines - https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ 
The changes are as follows 
* all src/* are moved to src/kepler_model so that `from kepler_model.X import Y` works (as guideline recommends)
* All executable python scripts are generated using `project.script`. These scripts essentially call the `kepler_model.<component>.run` function
* All sys.path.append (hacks) are removed
* missing `__init__.py` has been added to all package 
* dockerfile changes are made to support the modifications needed by above

Please note further cleanup / optimisations may be possible but can be done in a follow up patch.


vprashar2929:
@sthaha adding a description?

sunya-ch (approved):
Thank you for your hard effort. 
All PRs passes and this PR looks good to me. 
Some more explanation in the description would help for onboard the new approach for module calling.

Issue #324, opened 2024-07-24
Model training fails for Kepler release-0.7.11


vprashar2929:
### What happened?

When trying to train the model locally in case of Kepler with `release-0.7.11` is deployed it fails with the below trace
```
valid feature group:  [<FeatureGroup.CounterOnly: 3>, <FeatureGroup.BPFOnly: 5>, <FeatureGroup.BPFIRQ: 9>, <FeatureGroup.CounterIRQCombined: 7>]
/home/fserver/repos/kepler-model-server-vp/src/train/extractor/extractor.py:234: RuntimeWarning: Mean of empty slice.
  time_diff_values = df.reset_index()[[TIMESTAMP_COL]].diff().dropna().values.mean()
/home/fserver/.local/share/hatch/env/virtual/kepler-model-server/_v8KwWUN/kepler-model-server/lib/python3.10/site-packages/numpy/core/_methods.py:189: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
/home/fserver/repos/kepler-model-server-vp/src/train/extractor/extractor.py:234: RuntimeWarning: Mean of empty slice.
  time_diff_values = df.reset_index()[[TIMESTAMP_COL]].diff().dropna().values.mean()
/home/fserver/.local/share/hatch/env/virtual/kepler-model-server/_v8KwWUN/kepler-model-server/lib/python3.10/site-packages/numpy/core/_methods.py:189: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
/home/fserver/repos/kepler-model-server-vp/src/train/extractor/extractor.py:234: RuntimeWarning: Mean of empty slice.
  time_diff_values = df.reset_index()[[TIMESTAMP_COL]].diff().dropna().values.mean()
/home/fserver/.local/share/hatch/env/virtual/kepler-model-server/_v8KwWUN/kepler-model-server/lib/python3.10/site-packages/numpy/core/_methods.py:189: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
/home/fserver/repos/kepler-model-server-vp/src/train/extractor/extractor.py:234: RuntimeWarning: Mean of empty slice.
  time_diff_values = df.reset_index()[[TIMESTAMP_COL]].diff().dropna().values.mean()
[...]

PR #325, opened 2024-07-25
bump kepler version 0.7.11

sunya-ch:
Bumping kepler version to v0.7.11 with compatibility fixes.

If this PR passes and merged, it should solve https://github.com/sustainable-computing-io/kepler-model-server/issues/324.


.github/workflows/tekton-test.yml
```diff
@@ -154,7 +154,7 @@ jobs:
             - name: MACHINE_ID
               value: test
             - name: ENERGY_SOURCE
-              value: intel_rapl
+              value: trained_power_model
```
sthaha, line 157: isn't this wrong? doesn't it mean we are using estimator even in tekton workflow?

sunya-ch: It is intentionally. Tekton test on GH action runs on VM and its source is trained_power_model.

sunya-ch:
The only failed test is the integration-test-with-exporter-and-db. The reason is that we have no power model of ec2 for `rapl-sysfs`. However, we need the kepler-model-server image to collect and retrain the model. 
I would like to have this PR merges with known failure of `integration-test-with-exporter-and-db` test.

sthaha (approved):

Issue #247, opened 2024-04-05
timely CI error due to DNS failed to resolve the service


sunya-ch:
### What happened?

We found CI error failed from time to time (rerun for multiple times helps it pass)

```
error: connection error: Post "http://kepler-model-server.kepler.svc.cluster.local:8100/model": dial tcp: lookup kepler-model-server.kepler.svc.cluster.local on 10.96.0.10:53: no such host (http://kepler-model-server.kepler.svc.cluster.local:8100/model))
```

```
Error from server (InternalError): error when creating "tasks/train-task.yaml": Internal error occurred: failed calling webhook "webhook.pipeline.tekton.dev": failed to call webhook: Post "[https://tekton-pipelines-webhook.tekton-pipelines.svc:443/defaulting?timeout=10s](https://tekton-pipelines-webhook.tekton-pipelines.svc/defaulting?timeout=10s)": dial tcp 10.96.111.114:443: connect: connection refused
```

### What did you expect to happen?

Investigate root cause and fix

### How can we reproduce it (as minimally and precisely as possible)?

Push PR

### Kepler image tag

<details>

</details>

### Deployment

- [ ] Model server
- [ ] Estimator
- [ ] Online trainer
- [ ] Offline trainer
- [ ] Profiler

### Kepler model server image tag if deployed

<details>

</details>

### Kepler estimator image tag if deployed

<details>

</details>

### Kepler online trainer image tag if deployed

<details>

</details>

### Kepler offline trainer image tag if deployed

<details>

</details>

### Kepler profiler image tag if deployed

<details>

</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

### Install tools

<details>

</details>

### Kepler deployment config

<details>

For on kubernetes:
```console
$ KEPLER_NAMESPACE=kepler

# provide kepler configmap
$ kubectl get configmap kepler-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler model server configmap if Kepler Model Server is deployed 
$ kubectl get configmap kepler-model-server-cfm -n ${KEPLER_NAMESPACE} 
# paste output here

# provide kepler deployment description
$ kubectl describe deployment kepler-exporter -n ${KEPLER_NAMESPACE} 
```

For standalone:
# put your Kepler command argument here

</details>

PR #370, opened 2024-08-15
feat: update select logic with spec similarity computation

sunya-ch:
Related to https://github.com/sustainable-computing-io/kepler-model-server/issues/216, this PR adds a logic to find a node_type from the spec information. 

The idea is:
1. find exact node_type that equals or covers the machine spec
2. if not found, compute similarity between the machine spec and each node_type and select the most similar node_type
3. if more than one candidates with the same score, find uncertainty of selection (frequency of selected type over all candidates) and select the most frequent one.

This PR also updates the machine spec discovery in estimator to align with the kepler PR https://github.com/sustainable-computing-io/kepler/pull/1684. The updated logic is to first check availability of mounted spec first. If not available, apply the discover function.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ PYTHON = python3.10
 
 DOCKERFILES_PATH := ./dockerfiles
 MODEL_PATH := ${PWD}/tests/models
+MACHINE_SPEC_PATH := ${PWD}/tests/data/machine_spec
 
 build:
 	$(CTR_CMD) build -t $(IMAGE) -f $(DOCKERFILES_PATH)/Dockerfile .
@@ -53,10 +54,15 @@ run-estimator:
 
 run-collector-client:
 	$(CTR_CMD) exec estimator /bin/bash -c \
-		"while [ ! -S "/tmp/estimator.sock" ]; do sleep 1; done; hatch test -vvv -s ./tests/estimator_power_request_test.py"
+		"while [ ! -S "/tmp/estimator.sock" ]; do \
+			sleep 1; \
+		done; \
+		hatch run test -vvv -s ./tests/estimator_power_request_test.py"
 
 clean-estimator:
-	$(CTR_CMD) stop estimator
+	@$(CTR_CMD) logs estimator
+	@$(CTR_CMD) stop estimator
+	@$(CTR_CMD) rm estimator || true
 
 test-estimator: run-estimator run-collector-client clean-estimator
 
@@ -76,7 +82,9 @@ run-estimator-client:
 		hatch run test -vvv -s ./tests/estimator_model_request_test.py
 
 clean-model-server:
+	@$(CTR_CMD) logs model-server
 	@$(CTR_CMD) stop model-server
+	@$(CTR_CMD) rm model-server || true
 
 test-model-server: \
 	run-model-server \
@@ -103,6 +111,42 @@ test-offline-trainer: \
 	run-offline-trainer \
 	run-offline-trainer-client \
 	clean-offline-trainer
+
+# test model server select
+create-container-net:
+	@$(CTR_CMD) network create kepler-model-server-test
+
+run-model-server-with-db:
+	$(CTR_CMD) run -d --platform linux/amd64 \
+		--network kepler-model-server-test \
+		-p 8100:8100 \
+		--name model-server $(TEST_IMAGE) \
+		model-server
+	while ! docker logs model-server 2>&1 | grep -q 'Running on all'; do \
+		echo "... waiting for model-server to serve";  sleep 5; \
+	done
+
+run-estimator-with-model-server:
+	$(CTR_CMD) run -d --platform linux/amd64 \
+		--network kepler-model-server-test \
+		-e "PYTHONUNBUFFERED=1" \
+		-e "MACHINE_ID=test" \
+		-v ${MACHINE_SPEC_PATH}:/etc/kepler/models/machine_spec \
+		-e "MODEL_SERVER_ENABLE=true" \
+		-e "MODEL_SERVER_URL=http://model-server:8100" \
+		--name estimator $(TEST_IMAGE) \
+		estimator
+
+clean-container-net:
+	@$(CTR_CMD) network rm kepler-model-server-test
+
+run-select-client:
+	$(CTR_CMD) exec model-server \
+		hatch run test -vvv -s ./tests/model_select_test.py
+
+test-model-server-select: create-container-net run-model-server-with-db run-select-client clean-model-server clean-container-net
+
+test-model-server-estimator-select: create-container-net run-model-server-with-db run-estimator-with-model-server run-collector-client clean-estimator clean-model-server clean-container-net
 
 test: \
 	build-test \
--- a/src/kepler_model/estimate/estimator.py
+++ b/src/kepler_model/estimate/estimator.py
@@ -70,7 +70,7 @@ def handle_request(data):
             current_trainer = loaded_model[output_type.name][power_request.energy_source].trainer_name
             request_trainer = current_trainer != power_request.trainer_name
             if request_trainer:
-                logger.info("try obtaining the requesting trainer {} (current: {})".format(power_request.trainer_name, current_trainer))
+                logger.info(f"try obtaining the requesting trainer {power_request.trainer_name} (current: {current_trainer})")
     if power_request.energy_source not in loaded_model[output_type.name] or request_trainer:
         output_path = get_download_output_path(download_path, power_request.energy_source, output_type)
         if not os.path.exists(output_path):
@@ -84,20 +84,20 @@ def handle_request(data):
                     logger.error(msg)
                     return {"powers": dict(), "msg": msg}
                 else:
-                    logger.info("load model from config: ", output_path)
+                    logger.info(f"load model from config: {output_path}")
             else:
-                logger.info("load model from model server: %s", output_path)
+                logger.info(f"load model from model server: {output_path}")
         loaded_item = load_downloaded_model(power_request.energy_source, output_type)
         if loaded_item is not None and loaded_item.estimator is not None:
             loaded_model[output_type.name][power_request.energy_source] = loaded_item
-            logger.info("set model {0} for {2} ({1})".format(loaded_item.model_name, output_type.name, power_request.energy_source))
+            logger.info(f"set model {loaded_item.model_name} for {power_request.energy_source} ({output_type.name})")
         # remove loaded model
         shutil.rmtree(output_path)
 
     model = loaded_model[output_type.name][power_request.energy_source]
     powers, msg = model.get_power(power_request.datapoint)
     if msg != "":
-        logger.info("{} fail to predict, removed: {}".format(model.model_name, msg))
+        logger.info(f"{model.model_name} fail to predict, removed: {msg}")
         if output_path != "" and os.path.exists(output_path):
             shutil.rmtree(output_path)
     return {"powers": powers, "msg": msg}
@@ -111,7 +111,7 @@ class EstimatorServer:
         s = self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         s.bind(self.socket_path)
         s.listen(1)
-        logger.info("started serving on {}".format(self.socket_path))
+        logger.info(f"started serving on {self.socket_path}")
         try:
             while True:
                 connection, _ = s.accept()
--- a/src/kepler_model/estimate/model_server_connector.py
+++ b/src/kepler_model/estimate/model_server_connector.py
@@ -8,22 +8,31 @@ import codecs
 from kepler_model.util.config import is_model_server_enabled, get_model_server_req_endpoint, get_model_server_list_endpoint, download_path
 from kepler_model.util.loader import get_download_output_path
 from kepler_model.util.train_types import ModelOutputType
+from kepler_model.train.profiler.node_type_index import discover_spec_values
 
+machine_spec_mount_path = "/etc/kepler/models/machine_spec"
+machine_id = os.getenv('MACHINE_ID', None)
 
-# discover_spec: determine node spec in json format (refer to NodeTypeSpec)
-def discover_spec():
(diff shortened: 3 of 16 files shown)


sunya-ch:
Note that need to update ComponentModelWeights in kepler regressor module to unmarshal also the model_name field.

Makefile
```diff
@@ -76,7 +82,9 @@ run-estimator-client:
 		hatch run test -vvv -s ./tests/estimator_model_request_test.py
 
 clean-model-server:
+	@$(CTR_CMD) logs model-server
 	@$(CTR_CMD) stop model-server
+	@$(CTR_CMD) rm model-server || true
```
sthaha, line 87: Aren't we already adding --rm to the `run` ? I think we just need to `stop` and the container will be automatically deleted.

dockerfiles/Dockerfile
```diff
@@ -6,7 +6,7 @@ ENV PYTHONPATH=/kepler_model
 COPY pyproject.toml .
 COPY README.md .
 COPY cmd/ cmd/
-COPY src/ src/
+COPY src/kepler_model src/kepler_model
```
sthaha, line 9: lets keep the same and let `dockerignore ignore any files.

dockerfiles/Dockerfile.test
```diff
@@ -14,7 +14,7 @@ ENV PYTHONPATH=/kepler_model
 COPY pyproject.toml .
 COPY README.md .
 COPY cmd/ cmd/
-COPY src/ src/
+COPY src/kepler_model src/kepler_model
```
sthaha, line 17: same as above :)

src/kepler_model/estimate/estimator.py
```diff
@@ -85,7 +85,7 @@ def handle_request(data):
                     logger.error(msg)
                     return {"powers": dict(), "msg": msg}
                 else:
-                    logger.info("load model from config: ", output_path)
+                    logger.info("load model from config: {}".format(output_path))
```
sthaha, line 88:
```suggestion
                    logger.info("load model from config:  %s", output_path)
```

src/kepler_model/estimate/model_server_connector.py
```diff
@@ -8,22 +8,31 @@
...
 from kepler_model.util.train_types import ModelOutputType
+from kepler_model.train.profiler.node_type_index import discover_spec_values
 
+machine_spec_mount_path = "/etc/machine_spec"
 
-# discover_spec: determine node spec in json format (refer to NodeTypeSpec)
-def discover_spec():
-    import psutil
+# get_spec_values: determine node spec in json format (refer to NodeTypeSpec)
+def get_spec_values():
```
sthaha, line 16:
```suggestion
def get_spec_values(machine_id : str|None):
```
Let the caller pass the env var, this makes the function unit testable as well .

src/kepler_model/estimate/model_server_connector.py
```diff
@@ -62,11 +71,18 @@ def make_request(power_request):
...
         return dict()
     try:
-        response = requests.get(get_model_server_list_endpoint())
+        endpoint = get_model_server_list_endpoint()
+        if energy_source is not None:
+            endpoint = f"{endpoint}?source={energy_source}"
+            if node_type is not None:
+                endpoint = f"{endpoint}&type={node_type}"
+        elif node_type is not None:
+            endpoint = f"{endpoint}?type={node_type}"
```
sthaha, line 84:
```suggestion
        if energy_source: # also ignore empty str
            endpoint = f"{endpoint}?source={energy_source}"
            
        if node_type:
            endpoint = f"{endpoint}?type={node_type}"
```

src/kepler_model/server/model_server.py
```diff
@@ -84,11 +100,11 @@ def select_best_model(spec, valid_groupath, filters, energy_source, pipeline_nam
             print("no large candidates, select from all availables")
             candidates = model_names
     for model_name in candidates:
-        model_savepath = os.path.join(valid_groupath, model_name)
+        model_savepath = os.path.join(valid_grouppath, model_name)
         metadata = load_json(model_savepath, METADATA_FILENAME)
         if metadata is None or not is_valid_model(metadata, filters) or ERROR_KEY not in metadata:
             # invalid metadata
-            print("invalid", is_valid_model(metadata, filters), metadata)
+            print(f"skip invalid {model_name} candidate - model validity: {is_valid_model(metadata, filters)} metadata: {metadata}")
```
sthaha, line 107: can we use `logging` to log these ?

src/kepler_model/train/profiler/node_type_index.py
```diff
@@ -137,6 +158,37 @@ def cover(self, compare_spec):
...
+                if compare_spec.attrs[attr] is not None:
+                    similarity = compute_similarity(self.attrs[attr], compare_spec.attrs[attr])
+            if debug:
+                print(attr, self.attrs[attr], compare_spec.attrs[attr], similarity, get_similarity_weight(attr))
+            assert similarity >= 0, "similarity must be >= 0, {} ({})".format(attr, similarity)
+            assert similarity <= 1, "similarity must be <= 1, {} ({})".format(attr, similarity)
+            total_similarity += (similarity*get_similarity_weight(attr))
+        if total_similarity > 1:
+            total_similarity = 1
+        return total_similarity
```
sthaha, line 190: Can we move this to a test? Is it possible to write a unit-test for this?

tests/README.md
```diff
@@ -250,6 +250,23 @@ Optional arguments:
...
+## Server API
+The test is for testing server model selection API.
+
+Requirements:
+- server running
+
+    ```bash
+    # create new model folder
+    mkdir -p $(pwd)/db-models
+    MODEL_PATH=$(pwd)/db-models python ../src/server/model_server.py
```
sthaha, line 262:
```suggestion
    MODEL_PATH=$(pwd)/db-models model-server
```

Makefile
sunya-ch: It should be but it fails to remove sometimes before starting the new test which also need model-server container. It is the reason that I add rm line with || true.

sunya-ch:
@sthaha thank you so much for the reviews. Made an update.
Also need the following PR in kepler to be merged first to support model_name attribute in weight file: 
- [ ] https://github.com/sustainable-computing-io/kepler/pull/1699.

rootfs:
1699 is merged, shall we merge this one now?

rootfs (approved):

src/kepler_model/estimate/model_server_connector.py
```diff
@@ -8,22 +8,31 @@
 from kepler_model.util.config import is_model_server_enabled, get_model_server_req_endpoint, get_model_server_list_endpoint, download_path
 from kepler_model.util.loader import get_download_output_path
 from kepler_model.util.train_types import ModelOutputType
+from kepler_model.train.profiler.node_type_index import discover_spec_values
 
+machine_spec_mount_path = "/etc/machine_spec"
```
sthaha, line 13: can we keep all model-server related settings under `/etc/kepler/models/` ?

sunya-ch:
> 1699 is merged, shall we merge this one now?

I think we could but there are following things might be good to resolve first.
(i) I would like to have all green on test. I need some rerun the failed test due to the not-fixed-yet but known issue as raised in https://github.com/sustainable-computing-io/kepler-model-server/issues/247.
(ii) @sthaha may you check your comments are all resolved?
(iii) it might be good to see whether the compatibility is solved with the latest Kepler by having the following CI PR merged first.

- [ ] https://github.com/sustainable-computing-io/kepler-model-server/pull/374

src/kepler_model/estimate/model_server_connector.py
```diff
@@ -62,16 +71,22 @@ def make_request(power_request):
...
     try:
-        response = requests.get(get_model_server_list_endpoint())
+        endpoint = get_model_server_list_endpoint()
+        if energy_source:
+            endpoint = f"{endpoint}?source={energy_source}"
+            if node_type:
+                endpoint = f"{endpoint}&type={node_type}"
+        elif node_type:
+            endpoint = f"{endpoint}?type={node_type}"
+        response = requests.get(endpoint)
```
sthaha, line 85: How about ?

```suggestion
        endpoint = get_model_server_list_endpoint()
        params= {}
        if energy_source:
            params["source"] = energy_source
        if node_type:
           params["type"] = node_type
        response = requests.get(endpoint, params=params)
```

src/kepler_model/util/loader.py
```diff
@@ -60,7 +60,7 @@ def load_json(path, name):
         with open(filepath) as f:
             res = json.load(f)
         return res
-    except Exception as err:
+    except:
```
sthaha, line 63: Let use `Exception` at least since catch-all even masks typo in code.

src/kepler_model/server/model_server.py
```diff
@@ -253,15 +304,20 @@ def load_init_pipeline():
...
 
 
 def run():
-    load_init_pipeline()
+    logger.info("Run model-server")
+    try:
+        load_init_pipeline()
+    except Exception as e:
+        logger.error(f"failed to load initial pipeline {e}")
+        return
```
sthaha, line 320:
```suggestion
        return 1
```

src/kepler_model/estimate/model_server_connector.py
sunya-ch: Okay. Just to note that will also update in https://github.com/sustainable-computing-io/kepler/pull/1684 to the same path.

src/kepler_model/estimate/model_server_connector.py
```diff
@@ -8,22 +8,31 @@
...
-
-node_spec = discover_spec()
+node_spec = None
 
 
 def make_model_request(power_request):
+    global node_spec
+    if node_spec is None:
+         node_spec = get_spec_values(machine_id)
+         print(f"Node spec: {node_spec}")
```
sthaha, line 35: lets make use of `logging` to log or `click.echo` to print to console. This allows us (and `ruff`) to weed out debug messages from intentional console output.

src/kepler_model/train/profiler/node_type_index.py
```diff
@@ -183,13 +233,34 @@ def index_train_machine(self, machine_id, new_spec):
         self.node_type_index[covered_index].add_member(machine_id)
         return covered_index
 
-    def get_node_type(self, compare_spec):
+    def get_node_type(self, in_spec, loose_search=False, debug=False):
```
sthaha, line 236: this function is a bit too complex. Is it possible for us
* to break it down into smaller units
* write a unit-test ?

src/kepler_model/util/loader.py
```diff
@@ -88,6 +88,16 @@ def load_remote_pkl(url_path):
...
 
+def load_remote_json(url_path):
+    if ".json" not in url_path:
+        url_path = url_path + ".json"
+    try:        
+        response = urlopen(url_path)
+        response_data = response.read().decode('utf-8')
+        json_data = json.loads(response_data)
+        return json_data
+    except:
```
sthaha, line 99: lets catch `except Exception` so that typos in code are not masked.

src/kepler_model/util/similarity.py
```diff
@@ -0,0 +1,89 @@
...
+    NodeAttribute.FREQ: 0.01,
+}
+
+similarity_total_weight = sum(similarity_reference.values())
+
+def get_similarity_weight(attr):
+    return similarity_reference[attr]/similarity_total_weight
+
+def compute_jaccard_similarity(str1, str2):
+    if str1.lower() == str2.lower():
```
sthaha, line 27:
```suggestion
    if len(str1) == 0 or len(str2)==0:
       return 0

    if str1.lower() == str2.lower():
```

src/kepler_model/util/similarity.py
```diff
@@ -0,0 +1,89 @@
...
+        return 1
+    set1 = set(str1.lower())  # Convert to lowercase for case-insensitive comparison
+    set2 = set(str2.lower())
+
+    intersection = len(set1.intersection(set2))
+    union = len(set1.union(set2))
+
+    similarity = intersection / union if union != 0 else 0
+    return similarity/2
```
sthaha, line 36:
```suggestion
    similarity = intersection / union * 0.5
    return similarity
```
Given that `union` can't be 0 because of len() != 0 check above.

src/kepler_model/util/similarity.py
```diff
@@ -0,0 +1,89 @@
...
+    NodeAttribute.MEMORY: 0.8,
+    NodeAttribute.FREQ: 0.01,
+}
+
+similarity_total_weight = sum(similarity_reference.values())
+
+def get_similarity_weight(attr):
+    return similarity_reference[attr]/similarity_total_weight
+
+def compute_jaccard_similarity(str1, str2):
```
sthaha, line 26: how about we write a unit-test for this?

sthaha, line 26:
```suggestion
def compute_jaccard_similarity(str1 :str, str2: str) -> float:
```

src/kepler_model/util/similarity.py
```diff
@@ -0,0 +1,89 @@
...
+    set1 = set(str1.lower())  # Convert to lowercase for case-insensitive comparison
+    set2 = set(str2.lower())
+
+    intersection = len(set1.intersection(set2))
+    union = len(set1.union(set2))
+
+    similarity = intersection / union if union != 0 else 0
+    return similarity/2
+
+def compute_similarity(base, cmp):
```
sthaha, line 38:
```suggestion
def compute_similarity(base : float, cmp: float) -> float:
```

src/kepler_model/util/similarity.py
```diff
@@ -0,0 +1,89 @@
...
+    base = float(base)
+    cmp = float(cmp)
+    if base == 0 and cmp == 0:
+        diff_ratio = abs(cmp-base)
```
sthaha, line 42: If both are 0, isn't the diff-ratio already 0?

PR #375, opened 2024-08-19
fix(estimator): use click to handle log-level

sthaha:


sunya-ch (approved):
LGTM. Shall we also apply this for model_server as well?

sthaha:
> Shall we also apply this for model_server as well?

Definitely! I will submit another PR for that :) 
I also apply the same of kepler-model / aka cmd/main.py

PR #380, opened 2024-08-20
fix(estimator): crash when logging

sthaha:


sunya-ch:
Do we have the same issue here: https://github.com/sustainable-computing-io/kepler-model-server/blob/36d61038d278745dd456039d708c8a9218aa9859/src/kepler_model/estimate/estimator.py#L125?

sunya-ch (approved):
/LGTM Thanks!

PR #381, opened 2024-08-20
Revert "feat: update select logic with spec similarity computation"

sthaha:
Reverts sustainable-computing-io/kepler-model-server#370


sunya-ch (approved):

sunya-ch:
Will get back later after release 0.7.11-2.

PR #386, opened 2024-08-20
fix: ignore whitespaces in MODEL_CONFIG file

sthaha:
This commit trims whitespaces including newlines when reading the MODEL_CONFIG file.


sunya-ch (approved):
/lgtm

PR #392, opened 2024-08-21
feat: get type in /best-models API and pass source and type from estimator

sunya-ch:
This PR adds variable `type` to the /best-models API so that it can list the model by filtering the node_type.
This is required by the new feature to select model based on node_type and spec.

### DIFF
--- a/src/kepler_model/server/model_server.py
+++ b/src/kepler_model/server/model_server.py
@@ -189,6 +189,7 @@ def get_available_models():
     fg = request.args.get("fg")
     ot = request.args.get("ot")
     energy_source = request.args.get("source")
+    node_type = request.args.get("type")
     filter = request.args.get("filter")
 
     try:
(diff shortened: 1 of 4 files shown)


src/kepler_model/server/model_server.py
```diff
@@ -189,6 +189,7 @@ def get_available_models():
     fg = request.args.get("fg")
     ot = request.args.get("ot")
     energy_source = request.args.get("source")
+    node_type = request.args.get("type")
```
sthaha, line 192: Just a thought ... `type` is too generic.  should we call this node-type itself ?

src/kepler_model/server/model_server.py
```diff
@@ -205,18 +206,25 @@ def get_available_models():
         if energy_source is None or "rapl" in energy_source:
             energy_source = "rapl-sysfs"
 
+        if node_type is None:
+            node_type = any_node_type
+        else:
+            node_type = int(node_type)
```
sthaha, line 212: can we also add type hints to `select_best_model` to indicate that `node_type` must be `int` ?

sthaha (approved):

PR #398, opened 2024-08-22
feat: add --machine-spec arg to estimator and get_machine_spec

sunya-ch:
This PR updates kepler-model-server estimator with the same behavior as https://github.com/sustainable-computing-io/kepler/pull/1684.

1. pass --machine-spec file via command arguments

with spec file:
```
INFO:kepler_model.estimate.estimator:clean socket
INFO:kepler_model.estimate.estimator:initialize EstimatorServer with spec={'processor': 'intel_xeon_platinum_8259cl', 'cores': 96, 'chips': 2, 'memory': 377, 'frequency': 3500}
INFO:kepler_model.estimate.estimator:started serving on /tmp/estimator.sock
```
without spec file (auto-generated):
```
INFO:kepler_model.estimate.estimator:clean socket
INFO:kepler_model.estimate.estimator:initialize EstimatorServer with spec={'vendor': 'apple', 'processor': '', 'cores': 8, 'chips': 1, 'memory': 14, 'threads_per_core': 1}
INFO:kepler_model.estimate.estimator:started serving on /tmp/estimator.sock
```
2. add get_machine_spec function
3. add attr_has_value function

### DIFF
--- a/src/kepler_model/estimate/estimator.py
+++ b/src/kepler_model/estimate/estimator.py
@@ -12,6 +12,7 @@ import pandas as pd
 from kepler_model.estimate.archived_model import get_achived_model
 from kepler_model.estimate.model.model import load_downloaded_model
 from kepler_model.estimate.model_server_connector import is_model_server_enabled, make_request
+from kepler_model.train.profiler.node_type_index import get_machine_spec
 from kepler_model.util.config import SERVE_SOCKET, download_path, set_env_from_model_config
 from kepler_model.util.loader import get_download_output_path
 from kepler_model.util.train_types import ModelOutputType, convert_enery_source, is_output_type_supported
@@ -39,11 +40,9 @@ class PowerRequest:
 ###############################################
 # serve
 
-
 loaded_model = dict()
 
-
-def handle_request(data: str) -> dict:
+def handle_request(data: str, machine_spec=None) -> dict:
     try:
         power_request = json.loads(data, object_hook=lambda d: PowerRequest(**d))
     except Exception as e:
@@ -78,7 +77,7 @@ def handle_request(data: str) -> dict:
             shutil.rmtree(output_path)
         if not os.path.exists(output_path):
             # try connecting to model server
-            output_path = make_request(power_request)
+            output_path = make_request(power_request, machine_spec)
             if output_path is None:
                 # find from config
                 output_path = get_achived_model(power_request)
@@ -107,8 +106,10 @@ def handle_request(data: str) -> dict:
 
 
 class EstimatorServer:
-    def __init__(self, socket_path):
+    def __init__(self, socket_path, machine_spec):
         self.socket_path = socket_path
+        self.machine_spec = machine_spec
+        logger.info(f"initialize EstimatorServer with spec={machine_spec}")
 
     def start(self):
         s = self.socket = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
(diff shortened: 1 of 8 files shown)


src/kepler_model/estimate/estimator.py
```diff
@@ -160,15 +161,22 @@ def sig_handler(signum, frame) -> None:
     default="info",
     required=False,
 )
-def run(log_level: str):
+@click.option(
+    "--machine-spec",
+    type=str,
```
sthaha, line 166: this can be a path  see usage: https://github.com/sustainable-computing-io/kepler/blob/main/e2e/tools/validator/src/validator/cli/__init__.py#L243 

and click will check if the path exists etc ..

tests/model_server_test.py
```diff
@@ -43,8 +45,35 @@ def get_models():
...
+        assert attr_has_value(attrs, NodeAttribute.PROCESSOR) == value
+    for tc, value in numerical_test_cases.items():
+        attrs[NodeAttribute.CORES] = tc
+        assert attr_has_value(attrs, NodeAttribute.CORES) == value
+
+def run_model_server_unit_tests():
+    test_attr_has_value()
 
 if __name__ == "__main__":
+    run_model_server_unit_tests()
```
sthaha, line 76: could we move all unit-tests to a `test_<component>.py` file that pytest can run?

sthaha, line 76:
```suggestion
if __name__ == "__main__":
```

src/kepler_model/estimate/model_server_connector.py
```diff
@@ -54,10 +48,10 @@ def unpack(energy_source, output_type, response, replace=True):
...
 
-def make_request(power_request):
+def make_request(power_request, machine_spec):
     if not is_model_server_enabled():
         return None
-    model_request = make_model_request(power_request)
+    model_request = make_model_request(power_request, machine_spec)
     output_type = ModelOutputType[power_request.output_type]
     try:
         response = requests.post(get_model_server_req_endpoint(), json=model_request)
```
sthaha, line 57: Unrelated but .. FYI since I see that we are using POST here. POSTs are meant to make some changes to the server. 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST these are unsafe methods and thus won't be  cached by any intermediaries.

src/kepler_model/train/profiler/node_type_index.py
```diff
@@ -34,6 +39,8 @@ def rename(name: str) -> str:
 
 
 def format_processor(processor):
+    if len(processor) < 2: # brand_raw is set to "-" on some machine
```
sthaha, line 42: minor: do you want to call it `processors` to indicate plural and also `type hint` as `list[str]` ?

src/kepler_model/train/profiler/node_type_index.py
```diff
@@ -86,9 +116,18 @@ def load_node_type_spec(node_type_index_json):
             node_type_spec_index[int(index)] = spec
     return node_type_spec_index
 
-
 no_data = None
 
+def attr_has_value(attrs: dict, key: NodeAttribute) -> bool:
```
sthaha, line 121: nice!

sthaha (approved):

src/kepler_model/train/profiler/node_type_index.py
sunya-ch: processor is string. This line to avoid the case that I found the processor cannot be retrieved since its value is "-".

src/kepler_model/estimate/model_server_connector.py
sunya-ch: Oh I see. Yes, agree that we should change to GET. But this needs to change in both estimator and kepler side. Let's do that later in a graceful way.

src/kepler_model/train/profiler/node_type_index.py
sthaha: how about we type hint :str. ?

PR #444, opened 2024-09-12
fix(estimator): logging and machine-spec computation

sthaha:
This PR introduces the following changes

* improves logging to report file and line info `2024-09-12 05:49:15 ERROR loader.py:79: failed to load json metadata.json`
* removes unnecessary `cat ... |` when grepping for physical id in /proc/cpuinfo

### DIFF
--- a/src/kepler_model/train/profiler/node_type_index.py
+++ b/src/kepler_model/train/profiler/node_type_index.py
@@ -77,7 +77,7 @@ def discover_spec_values():
         vendor = format_vendor(cpu_info["vendor_id_raw"])
 
     cores = psutil.cpu_count(logical=True)
-    chips = max(1, int(subprocess.check_output('cat /proc/cpuinfo | grep "physical id" | sort -u | wc -l', shell=True)))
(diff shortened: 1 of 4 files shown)


sunya-ch (approved):
looks good to me. Thanks!

src/kepler_model/train/profiler/node_type_index.py
```diff
@@ -77,7 +77,7 @@ def discover_spec_values():
         vendor = format_vendor(cpu_info["vendor_id_raw"])
 
     cores = psutil.cpu_count(logical=True)
-    chips = max(1, int(subprocess.check_output('cat /proc/cpuinfo | grep "physical id" | sort -u | wc -l', shell=True)))
+    chips = max(1, int(subprocess.check_output('grep --count "physical id" /proc/cpuinfo', shell=True)))
```
sthaha, line 80: @sunya-ch  on a multi-cpu system do we not want to count all `physical id` s ?

sthaha: Reverting to just remove `cat`

sunya-ch: We do. max 1 is to avoid zero.

PR #456, opened 2024-09-19
feat: support --config-dir arg to point to the configuration directory

sthaha:
This PR add support for 
* passing path to config-dir using `--config-dir | -c` 
* allows MODEL_CONFIG to contain comments `#`


KaiyiLiu1234 (approved):
LGTM

PR #462, opened 2024-09-23
chore(pyproject): add an env to run jupyter lab

sthaha:
This allows us to easily bring up jupyter lab by running
```
hatch run lab:note [--any optional args]
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ authors = [
 
 classifiers = [
   "Programming Language :: Python",
-	"Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3",
   "Programming Language :: Python :: 3.10",
 ]
 dependencies = [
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -83,6 +83,27 @@ cov = [
...
+	"voila",
+	"ipywidgets",
+	# vim please
+	"jupyterlab-vim",
+
+	"beautifulsoup4",
+	# read parquet files
+	# "pyarrow",
+
+  # graphing
```
sunya-ch, line 99: tab?

pyproject.toml
```diff
@@ -83,6 +83,27 @@ cov = [
...
+	"jupyterlab",
+	"notebook",
+	"voila",
+	"ipywidgets",
+	# vim please
+	"jupyterlab-vim",
+
+	"beautifulsoup4",
+	# read parquet files
+	# "pyarrow",
```
sunya-ch, line 97: why comment this?

sunya-ch (approved):
/lgtm with some minor comments.

pyproject.toml
sthaha: We are not reading parquet yet but in case we switch to using those, this comes handy.

PR #464, opened 2024-09-23
feat(vm_metrics): Enabled vm metric use for local model training

KaiyiLiu1234:
Currently, the model server training setup uses baremetal metrics only. Included a "use_vm_metrics" field to enable model training with job='vm' metrics and kepler_vm energy sources. This will be useful set up for further testing.


src/kepler_model/cmd/main.py
```diff
@@ -1018,6 +1022,12 @@ def run():
     parser.add_argument(
         "--trainers", type=str, help="Specify trainer names for train_from_data command (use comma(,) as delimiter).", default="XgboostFitTrainer"
     )
+    parser.add_argument(
+        "--vm-train",
+        type=str,
```
sunya-ch, line 1027: why not bool?

src/kepler_model/train/extractor/extractor.py
```diff
@@ -160,6 +162,11 @@ def get_workload_feature_data(self, query_results, features):
             aggr_query_data = query_results[query].copy()
 
             if all(col in aggr_query_data.columns for col in container_id_cols):
+                if use_vm_metrics:
+                    aggr_query_data = aggr_query_data.loc[aggr_query_data["job"] == VM_JOB_NAME]
+                else:
+                    aggr_query_data = aggr_query_data.loc[aggr_query_data["job"] != VM_JOB_NAME]
```
sunya-ch, line 168: Don't think we need this else condition. To avoid unexpected job filtering on general case, I think we can filter only when enable use_vm_metrics.

src/kepler_model/util/prom_types.py
```diff
@@ -72,6 +77,10 @@ def energy_component_to_query(component):
     return f"{node_query_prefix}_{component}_{node_query_suffix}"
 
 
+def vm_energy_component_to_query(component):
```
sunya-ch, line 80: I don't think we should use the the kepler_vm for training. We should use kepler_node but with job="metal" label (or job != "vm"). kepler_vm is not from the power meter but from the ratio which considering fairness distribution. In the case the we have only single vm, this might not be different but I think we better use the number from power meter.

src/kepler_model/train/extractor/extractor.py
```diff
@@ -229,15 +236,20 @@ def get_system_feature_data(self, query_results, features):
...
-            query = energy_component_to_query(component)
+            if use_vm_metrics:
+                query = vm_energy_component_to_query(component)
+            else:
+                query = energy_component_to_query(component)
             if query not in query_results:
                 print(query, "not in", query_results)
                 return None
             aggr_query_data = query_results[query].copy()
+            if not use_vm_metrics:
```
sunya-ch, line 251: Could you please comment an explanation of this line?

src/kepler_model/train/extractor/extractor.py
KaiyiLiu1234: My concern is that if the filter is not added, job=vm metrics will also be pulled by the model server alongside baremetal metrics which is not a desirable result right?

src/kepler_model/train/extractor/extractor.py
KaiyiLiu1234: So if it is baremetal, I do not want it to use job=vm energy metric queries. We should only use baremetal metrics. Basically, I am making sure that the behavior of the extractor prior to my addition of vm metrics is the same as before (which is it will extract all the metrics with kepler_node_*). Same with in get_workload_feature_dataset.

src/kepler_model/train/pipeline.py
```diff
@@ -241,6 +249,7 @@ def NewPipeline(
     isolator=MinIdleIsolator(),
     target_energy_sources=PowerSourceMap.keys(),
     valid_feature_groups=FeatureGroups.keys(),
+    use_vm_metrics=False,
```
rootfs, line 252: can you update the offline_trainer.py to pick up this new param from e.g. env var?

src/kepler_model/train/extractor/extractor.py
sunya-ch: Just thinking that the label job is a common label that could be relabeled for example if we want to label the pod with the application job. However, I understand your points that we might want to get the metric of BM also for the case that we enable the VM. After some thought, I think we can keep it as you code.

src/kepler_model/train/extractor/extractor.py
sunya-ch: can we have a single splitter to split between vm_query_results and bm_query_results from a single query_result? It could be apply for the feature extract as well.  
Here I think it is fine to filter only for bm query since the vm query has `vm` word which is only available on bm. However, as I mentioned below, I still doubt to use vm value not the node value.

src/kepler_model/util/prom_types.py
KaiyiLiu1234: @rootfs thoughts? If I were to use kepler_node, I would need to find the process id for the virtual machine. If I recall, kepler_node with vm process id and kepler_vm are the same value or very similar.

src/kepler_model/train/extractor/extractor.py
KaiyiLiu1234: Good point. I will change that now

rootfs (approved):

src/kepler_model/util/prom_types.py
sunya-ch: I don't think vm process will be the same if the idle power is not zero since it will distribute evenly for all running process not only the vm process but I think kepler_vm and kepler_node should be the same if you have a single VM.

sthaha (changes requested):
IMHO, we should test if this works to yield better results  before we merge it. 
What say @sunya-ch @rootfs ?

KaiyiLiu1234:
> IMHO, we should test if this works to yield better results before we merge it. What say @sunya-ch @rootfs ?

This should be sufficient proof: https://github.com/sustainable-computing-io/kepler-metal-ci/blob/main/docs/analytics/vm-cpu-time-predict-vm-power.ipynb - I am going to replicate this in a separate extractor using process instead of container metrics.

src/kepler_model/train/extractor/extractor.py
KaiyiLiu1234: Can we merge this as is because the issue is a lot of the methods in DefaultExtractor which imo should be protected are being used directly (like get_workload_feature_data and get_power) in other functions. So I can't really create a single splitter in the DefaultExtractor.extract method. I can also close this because I am right now creating a new Extractor (DefaultVMExtractor) which will include this PR and the new Extractor. @sunya-ch

sunya-ch:
@sthaha I see the notebook and it looks good to me. Even though the total CPU time of process inside VM not smooth as it should be but it can still give a clue for modeling.

This is my understanding for the ways to build the model and predict VM's power.
- BM's total CPU time -> BM's total energy (without running any VM) - current
- VM's total CPU time -> BM's energy of VM (running VM) - this PR

@KaiyiLiu1234 It would be great if you could add the result of the current version for the comparison. I cannot find total CPU time on BM in the current artifact. Could you point me?

FYI,
I do some more plots from the notebook.
- CPU times are slightly different from BM's VM and VM's total (expect straight line if equal)
Screenshot 2024-09-27 at 10 45 00
- The difference in CPU time observed inside and outside seems not significantly affecting the power model.
Screenshot 2024-09-27 at 11 15 31
* Note that, any of the above two predictions not including the current approach which use BM's total energy and BM's total CPU time as I cannot find BM's total CPU time in the artifact.

KaiyiLiu1234:
@sunya-ch +1 on your comment. Will produce issues for the points you made regarding the setup of this and begin work on that. Regarding your question on total cpu time in BM, it's not included in the validator when saving. It should be in the prometheus snapshot. We can also include it in the validator for future runs.

sthaha (commented):
> I agree that train inside and test inside is better than train outside and test inside for VM 

I worry if we are again falling for the trap of relying on the training data. E.g. have we tried training using VM, then have a different sized vm to test against to see if the accuracy is different? This should we much easy to test now. 

I am removing my "Request Changes" in the interest of time

src/kepler_model/cmd/cmd_util.py
```diff
@@ -354,6 +354,7 @@ def get_pipeline(
     energy_sources,
     valid_feature_groups,
     replace_node_type=default_node_type,
+    use_vm_metrics=False,
```
sthaha, line 357: I think we should allow passing labels to be selected instead. 
E.g. `kepler-model train --labels='job="vm"' --labels='foo="bar"` 
and keep a default of none, i.e single kepler instance.

sthaha: @KaiyiLiu1234 could you address this in a separate PR ?

src/kepler_model/cmd/main.py
```diff
@@ -406,6 +406,7 @@ def train_from_data(args):
 - --energy-source : specify target energy sources (use comma(,) as delimiter) 
 - --thirdparty-metrics : specify list of third party metric to export (required only for ThirdParty feature group)
 - --id : specify machine ID 
+- --vm-train: specify whether to use vm feature and energy metrics for training - true: use vm feature metrics. Default is false
```
sthaha, line 409: As indicated earlier, this should be replace with a way to select the `feature-group` metrics and a way to select the "target" metrics.

src/kepler_model/train/extractor/extractor.py
sunya-ch: @KaiyiLiu1234 Thank you for your contribution on this PR. In my opinion, creating a new DefaultVMExtractor is more preferable.

sunya-ch:
> > I agree that train inside and test inside is better than train outside and test inside for VM
> 
> I worry if we are again falling for the trap of relying on the training data. E.g. have we tried training using VM, then have a different sized vm to test against to see if the accuracy is different? This should we much easy to test now.

Continuing from the above comment, I think we should separate extractor as @KaiyiLiu1234  proposed and we need VM configuration information (as additional metadata) as required for providing the model with that extractor.

KaiyiLiu1234:
@rootfs I will merge this for now. When the model server undergoes its improvements, we can change this anyway.

KaiyiLiu1234:
@rootfs Let's merge this for now. We can use this until the Model Server undergoes improvements in the future.

PR #533, opened 2024-10-28
chore: use kepler_vm metrics as energy source

rootfs:
When training models with `kepler_vm_package_joules_total` or similar, we use the `trained_power_model` as energy source, this fixes the issue of getting the proper energy from prom metrics. 

Tested with the following command:

```bash
hatch run python cmd/main.py train --vm-train -e trained_power_model
```

### DIFF
--- a/src/kepler_model/cmd/cmd_util.py
+++ b/src/kepler_model/cmd/cmd_util.py
@@ -7,6 +7,7 @@ from kepler_model.util.loader import default_node_type, get_pipeline_path, load_
 from kepler_model.util.prom_types import (
     SOURCE_COL,
     energy_component_to_query,
+    vm_energy_component_to_query,
     node_info_column,
     prom_responses_to_results,
 )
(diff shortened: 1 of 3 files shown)


KaiyiLiu1234 (approved):
lgtm. Let's give this a try on metal ci.

src/kepler_model/cmd/cmd_util.py
```diff
@@ -175,10 +176,14 @@ def get_validate_df(data_path, benchmark_filename, query_response):
                     item["total"] = filtered_df[query].max()
                     items += [item]
     energy_queries = [query for query in query_results.keys() if "_joules" in query]
-    print("Energy Queries: ", container_queries)
+    print("Energy Queries: ", energy_queries)
```
sthaha, line 179:
```suggestion
    logger.debug(f"Energy Queries: {energy_queries} ")
```

rootfs: logger is not imported in model server. Let's have this in another refactoring.

